PR #13, opened 2022-10-31
Make User Abstract Class

manya-mittal:
Please review the Abstract User Class that I coded

### DIFF
--- a/src/main/java/Entities/User.java
+++ b/src/main/java/Entities/User.java
@@ -7,9 +7,9 @@ public abstract class User {
     private final String username;
     private final String password;
     private final List<Playlist> playlistList;
-    public User(String u, String p, ArrayList<Playlist> playlists){
-        this.password = p;
-        this.username = u;
+    public User(String username, String password, ArrayList<Playlist> playlists){
+        this.username = username;
+        this.password = password;
         this.playlistList = playlists;
     }
     public String getUsername() {
--- a/src/main/java/Entities/User.java
+++ b/src/main/java/Entities/User.java
@@ -1,4 +1,20 @@
 package Entities;
 
-public interface User {
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class User {
+    private String Username;
+    private String password;
+    List<Playlist> playlistList = new ArrayList<Playlist>();
+
+    public String getUsername() {
+        return Username;
+    }
+    public String getPassword() {
+        return password;
+    }
+    public List<Playlist> getPlaylistList() {
+        return playlistList;
+    }
 }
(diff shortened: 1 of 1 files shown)


src/main/java/Entities/User.java
```diff
@@ -1,4 +1,20 @@
 package Entities;
 
-public interface User {
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class User {
+    private String Username;
+    private String password;
+    List<Playlist> playlistList = new ArrayList<Playlist>();
```
sarxhlwalker, line 9: instance variable should have a private/public attribute

sarxhlwalker (changes requested):
Please see comments.

src/main/java/Entities/User.java
```diff
@@ -1,4 +1,20 @@
 package Entities;
 
-public interface User {
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class User {
+    private String Username;
+    private String password;
```
chrstinalin, line 8: Capitalization on Username
Kind of redundant but you /could/ make this one line? 

private String username, password;

chrstinalin (changes requested):
Very minor issues, otherwise good

chrstinalin (dismissed):
Looks good to me :)

src/main/java/Entities/User.java
rafaelgacesa: One liners can often lead to code which is a lot more difficult to debug; remember that primarily you write code to be read by other developers (maybe many months after you create the initial version).

The compiler doesn’t need one liners - in fact the good C and C++ compilers for instance will remove temporary variables and optimize the code far better than you can by writing one-liners.

sarxhlwalker (dismissed):
Wow, gorgeous code!

src/main/java/Entities/User.java
manya-mittal: Thanks for your feedback Christina! I agree with Rafael, I think code in one line might be more difficult to read and understand later

src/main/java/Entities/User.java
```diff
@@ -1,4 +1,24 @@
 package Entities;
 
-public interface User {
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class User {
+    private final String username;
+    private final String password;
+    private final List<Playlist> playlistList;
```
SiqiZhang6, line 9: delete plz

SiqiZhang6 (changes requested):

rafaelgacesa (approved):
I did NOT ask to be here. However,  it was a pleasure to read. Thanks!

sarxhlwalker (approved):
Wow, gorgeous code! Please merge this PR so I can stop looking at it.

PR #14, opened 2022-10-31
modified playlist object and created playlist factory

NicholasMacasaet:
fixed playlist class and added a playlist factory class

### DIFF
--- a/src/main/java/Entities/PlaylistFactory.java
+++ b/src/main/java/Entities/PlaylistFactory.java
@@ -1,4 +1,7 @@
 package Entities;
 
 public class PlaylistFactory {
+    public Playlist createPlaylist(String name, User owner, boolean isPublic){
+        return new Playlist(name,owner,isPublic);
+    }
 }
--- a/src/main/java/Entities/PlaylistFactory.java
+++ b/src/main/java/Entities/PlaylistFactory.java
@@ -2,6 +2,6 @@ package Entities;
 
 public class PlaylistFactory {
     public Playlist createPlaylist(String name, User owner, boolean isPublic){
-        return new Playlist(name,owner,isPublic);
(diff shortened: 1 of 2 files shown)


rafaelgacesa (approved):
Thanks Nick! Love this code. 👨‍💻

Mistryr222 (approved):
Excellent changes, I literally did this with you :D

src/main/java/Entities/PlaylistFactory.java
```diff
@@ -1,4 +1,7 @@
 package Entities;
 
 public class PlaylistFactory {
+    public Playlist createPlaylist(String name, User owner, boolean isPublic){
+        return new Playlist(name,owner,isPublic);
```
sarxhlwalker, line 5: Please put spaces after the commas in the constructor call

sarxhlwalker (changes requested):
Please see comments.

sarxhlwalker (approved):
Gorgeous code!

src/main/java/Entities/PlaylistFactory.java
rafaelgacesa: In Java, white space does not matter. In fact, Java regards the following program as being identical to that of Figure 1.2.

`import turtles.*;public class BaseProgram extends TurtleProgram
{public void run(){}}`

Figure 1.2:
```
  1  import turtles.*;
  2  
  3  public class BaseProgram extends TurtleProgram {
  4      public void run() {
  5          // This is a simple baseline program.
  6      }
  7  }
```


Whitespaces (blank space) in your code are ignored by the compiler, which just skips over them while compiling the code.

rafaelgacesa (approved):
Sarah is mean 😢

Mistryr222 (approved):
Great!

PR #15, opened 2022-10-31
Completed input/output data, input/output boundary classes; example databases

rafaelgacesa:
- Edits in Nick's #14 entity class
   - Typos in comments
   - Added javadocs
   - Added getter for list of songs

- Created the communal input/output data files I was assigned
   - playlistInputData, songOutputData
   - includes javadocs

- Created the input/output boundaries for the recommend song use case

- Created example databases for artists, users, and playlist
   - for reference as I create the data access classes


NicholasMacasaet:
thanks, there may have been methods I didn't implement yet

NicholasMacasaet (approved):
thanks, there may have been methods and descriptions I didn't implement yet

src/main/java/Database/artists.csv
```diff
@@ -0,0 +1,2 @@
+Artist1,Password1,Artist Name 1,110 200 3
```
chrstinalin, line 1: This is optional but I feel like separating the song IDs with commas would be nice (a lot of other data is going to end up being formatted that way, it'd be nice to reuse code but we could always just make it so that it takes in what to splice by).

src/main/java/Database/users.csv
```diff
@@ -0,0 +1,2 @@
+User1,Password1,Playlist 1;Playlist 2
```
chrstinalin, line 1: Lol. See other comment in regards to that semicolon separator

src/main/java/Database/artists.csv
rafaelgacesa: While I don't disagree that it looks nicer, since I am using .csv files (comma separated values), that would add columns and break the parsing. 

Also, I've already written the parsing methods for these databases, just haven't pushed yet since they rely on dependencies that don't exist as of today.

src/main/java/Database/users.csv
rafaelgacesa: Same thing :(. The way I have my parser written right now is that it takes a line at a time, splits by commas, and then I further split the subdata if necessary (using their respective delimiter).

src/main/java/InputData/playlistInputData.java
```diff
@@ -1,4 +1,32 @@
 package InputData;
 
+import Entities.Song;
+
+import java.util.ArrayList;
```
chrstinalin, line 5: Note the change from ArrayList -> LinkedList within Playlist class

chrstinalin (changes requested):
Made several comments of varying importance.

src/main/java/InputData/playlistInputData.java
rafaelgacesa: Yep, at Rohan's request. Doesn't really matter which iterable we use up until it gets into a Playlist object anyways.

src/main/java/Database/artists.csv
chrstinalin: Wow I completely forgot about the comma part of 'comma separated values' -- my bad, I probably shouldn't be reviewing this at 12am lol. Sticking to semicolons would be nice in case the difference causes some issue later down the line but it seems unlikely

rafaelgacesa: Yep, honestly I should've been consistent with my sub-delimiters anyways. I'll change the spaces to semicolons tomorrow and re-request your review.

src/main/java/InputData/playlistInputData.java
chrstinalin: I figured, thought I'd point it out just in case though

rafaelgacesa:
I lied, fixed sub-delimiters (sorry for the re-review Nick lol).

chrstinalin (approved):
Hurray for consistency 🥳

sarxhlwalker (approved):
Nice comments. Nothing stands out to me.

manya-mittal:
These changes look great! The javadocs will be useful in the future.

manya-mittal (approved):
These changes look great! The javadocs will be useful in the future.

PR #16, opened 2022-10-31
base work + renaming

sarxhlwalker:
Default files before I start on the use case. Some changes: 
- renamed spaceOutputBoundary -> spacePlayedOutputBoundary (spaceAdd does not have an OB, afaik)
- renamed playSpace -> playSpaceInteractor in order to avoid duplicate naming (playSpace is a function) -- assuming I've understood clean architecture correctly, I would recommend people do this for their use cases
-  renamed spaceInputBoundary -> spaceAddInputBoundary because we have two InputBoundaries for space functions and an arbitrary spaceIB is confusing

### DIFF
--- a/src/main/java/Controller/spaceController.java
+++ b/src/main/java/Controller/spaceController.java
@@ -1,4 +1,24 @@
 package Controller;
 
+import InputBoundary.playSpaceInputBoundary;
+import InputBoundary.spaceAddInputBoundary;
+import InputData.spaceInputData;
+
 public class spaceController {
+    private final spaceAddInputBoundary spaceAddInteractor;
+    private final playSpaceInputBoundary spacePlayInteractor;
+
+    public spaceController(spaceAddInputBoundary spaceAddInteractor, playSpaceInputBoundary spacePlayInteractor){
+        this.spaceAddInteractor = spaceAddInteractor;
+        this.spacePlayInteractor = spacePlayInteractor;
+    }
+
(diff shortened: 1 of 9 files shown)


rafaelgacesa (approved):
Good idea w/ the input/output boundary naming - I was really confused making them but just followed the CRC doc.

Wouldn't worry too much about the use case name, I don't think clean architecture has any concerns on class names as long as they're unambiguous? Not sure though and it's kinda whatever

Side note: you won't need the output data as a instance variable in your interactor, it's going to be created and sent to the presenter in the same method so storing it is pointless. approving this anyways since there's nothing really wrong with that, it just won't be accessed.

rafaelgacesa:
> Good idea w/ the input/output boundary naming - I was really confused making them but just followed the CRC doc.
> 
> Wouldn't worry too much about the use case name, I don't think clean architecture has any concerns on class names as long as they're unambiguous? Not sure though and it's kinda whatever
> 
> Side note: you won't need the output data as a instance variable in your interactor, it's going to be created and sent to the presenter in the same method so storing it is pointless. approving this anyways since there's nothing really wrong with that, it just won't be accessed.

Forgot to mention javadocs here too, we do need them (in the rubric) but I know some people like to comment later.

Mistryr222 (approved):
Thank you Sarah, very cool!

SiqiZhang6 (approved):
looks great

PR #17, opened 2022-11-01
Suggestion: remove Factory design pattern where it does not make sense

rafaelgacesa:
The Factory design pattern is ideal where we have multiple subclasses of one object, and we don't want to rely on strict dependencies to each subclass; instead we can create a factory and allow the sub-classes to decide what type of object to make.

We have Factories for Songs, Playlists, and Spaces, all of which are concrete classes; therefore I suggest that we remove these Factories as they do not make sense there.

See more [here](https://refactoring.guru/design-patterns/factory-method).


sarxhlwalker (approved):
IIRC we talked about this in tutorial today, but I think we should only remove them if we're certain we will not be implementing some of the additional features we scrapped in simplifying our project. For ex. if we ever decide to have collaborative playlists again, a playlist factory might be beneficial? (I don't remember our original plan for that, whether they were subclasses or a boolean instance variable.) I guess that mostly depends on what Nick thinks. 

However, I'm much more open to removing SongFactory and SpaceFactory -- I'm not sure what subclasses of Songs we'd need, and I think if we were to complicate our Space system if we had time, I'd prefer to work on the queuing function rather than implementing different spaces. (Different spaces would very much complicate the UI process, anyhow, which might be too much depending on however much extra time we theoretically have.)

rafaelgacesa:
True, collaborative playlists could require a subclass, but it's all up to implementation anyways (aka it's all up to Nick lol).

We already have that isPublic flag in the Playlist entity (which doesn't look like it'll be used in our current implementation plan), if we wanted to shoehorn in collaboration while still removing the Factory we could just allow Playlists with isPublic true to be edited by anyone.

chrstinalin (approved):
Removing Song makes sense to me -- in regards to Sarah's comment, I figure we could always just add it back if it comes down to it.

NicholasMacasaet (approved):
Will remove factory design for playlist when able to, thanks for the notice raf

NicholasMacasaet:
For collaborative playlists the overall implementation would be identical to a regular playlist object, so it wouldn't make sense to create an entirely new subclass with the same implementation when we could just have logic in the code that checks if the user is within the list of collaborators. And if so, then allow them to edit the playlist

PR #18, opened 2022-11-01
Song Class Implementation

chrstinalin:
What it says on the tin. Going with typecasting to a BufferedImage for the cover, which seems to play well with Swing.

### DIFF
--- a/src/main/java/Entities/Song.java
+++ b/src/main/java/Entities/Song.java
@@ -13,7 +13,17 @@ public class Song {
     private final boolean isExplicit;
     private final BufferedImage cover;
 
-
+    /**
+     *
+     * @param id Unique identifier of the song.
+     * @param name Name of the song.
+     * @param artistList List of contributing artists.
+     * @param length Length of the song, in seconds.
+     * @param genre Type of song.
+     * @param fileLocation Directory path to audio file.
+     * @param isExplicit Whether the song is explicit or not.
+     * @param cover Cover image of the song.
+     */
     public Song(int id, String name, List artistList, int length, String genre,
                 String fileLocation, boolean isExplicit, BufferedImage cover){
         this.id = id;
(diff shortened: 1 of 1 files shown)


rafaelgacesa (approved):
Class looks good, needs javadoc comments at some point though (see rubric). 

I'm approving it anyways since we need the class ASAP and I know some people prefer to add comments all at once later.

src/main/java/Entities/Song.java
```diff
@@ -1,4 +1,60 @@
...
+
+    public List getArtistList(){
+        return this.artistList;
+    }
+
+    public int getLength(){
+        return this.length;
+    }
+
+    public String getGenre(){
```
sarxhlwalker, line 45: I don't believe we need getGenre() anymore, as we're not having genre-related spaces. (Unless playlists require genre, but I don't think they do.)

src/main/java/Entities/Song.java
```diff
@@ -1,4 +1,60 @@
...
+        this.name = name;
+        this.artistList = artistList;
+        this.length = length;
+        this.genre = genre;
+        this.fileLocation = fileLocation;
+        this.isExplicit = isExplicit;
+        this.cover = cover;
+    }
+
+    public int getID(){
```
sarxhlwalker, line 29: Could you add a comment in the code (or add a brief response to this comment) explaining what the id is for? I assume database stuff, but I'm wondering whether I'd should/have to make use of the id variable.

sarxhlwalker (changes requested):
See comments.

src/main/java/Entities/Song.java
rafaelgacesa: Hi. Pls don't remove genre info, I use it for recommendations.

chrstinalin: I figured it'd be nice as a label next to songs also (like how length isn't really needed, either). You guys might not end up displaying it but I figure its better to have the option.

sarxhlwalker: Oops! Never mind then.

sarxhlwalker (approved):

src/main/java/Entities/Song.java
chrstinalin: Yeah I'll do it ASAP, You'd likely have to use the ID by the way, its the unique identifier for a given song rather than using its actual name (see: different songs with the same name in practice)

rafaelgacesa (approved):
Not sure if javadoc for constructor is usually done but comments ✅

rafaelgacesa:
Breaking the rules and force merging this. Sue me.
- people are _busy_
- there's nothing wrong with this code
- sitting around waiting for someone to glance over it and press approve before I can do any of my own work is just silly

PR #20, opened 2022-11-03
playlistLibrary, userList, interfaces, and other bits and pieces

rafaelgacesa:
Designed and created both the playlist and user database related classes and files.  Note that the version of `userList` in this pull is incomplete, as the required `User`, `ArtistUser`, `RegularUser`, and their respective factory classes do not exist yet. However, it is necessary to include it for all tests to pass, and so others can build their parts of the project using this code assuming the implementation is complete.

I’ve requested everyone to review because this is a **very** significant piece of the project. Everyone will interact with at least one of the two interfaces in the pull so it’s important to read them over. 

### Some notes:
- These databases are based on the Singleton design pattern, explained [here ](https://refactoring.guru/design-patterns/singleton)
    - in short, this means that they have private constructors, so only one instance can be created and accessed using the respective static `getInstance()` method
    - this design pattern is ideal for databases as having more than one database object writing to and from files would be catastrophic
- There was a technical violation of clean architecture in that the concrete database classes (`playlistLibrary`, `userLibrary`) rely on entities, but are in the frameworks & drivers layer
    - This was fixed in https://github.com/CSC207-2022F-UofT/course-project-team-communify/pull/20/commits/470c34600668c3c0048f2474c99d87f9506e5d06
- The `playlistLibrary` class relies on a method from `songLibrary` that doesn’t exist yet: `getSong(int id)`
    - I created a dummy implementation for now with a TODO
    - @chrstinalin if you have a different idea for how `songLibrary` is going to look please let me know ASAP
- The data access interfaces are the classes we’ll be using to communicate with the databases from use cases

### Other changes in this pull:
- Updated the structure of the .csv database files as I realized that some playlists may have identical names, so they need unique identifiers
- Added a unique id field to the `Playlist` entity
    - Constructor parameter list is getting a bit unwieldy, this may be an unavoidable code smell
- Created a very basic `RegularUser` class just to allow me to write some tests
- Created some unit tests for the playlist database class to get the ball rolling on those :)
    - I know we’re all probably dreading making unit tests but they’re a significant part of the rubric
    - **There should be unit tests for all use cases at the bare minimum**
- Updated `build.gradle` so that unit tests actually work
    - That was a very painful 20 minutes of debugging

### TL;DR: 
I like **bold**, unit tests are important, please read at least _playlistAccessInterface.java_ and _userAccessInterface.java_.

### DIFF
--- a/src/main/java/Database/playlistAccessInterface.java
+++ b/src/main/java/Database/playlistAccessInterface.java
@@ -1,4 +1,34 @@
 package Database;
 
+import Entities.Playlist;
+
+import java.util.Collection;
+
 public interface playlistAccessInterface {
+
+    /**
+     * @return ArrayList of all existing Playlists
+     */
+    public Collection<Playlist> getPlaylists();
+
+    /**
+     * Saves a new playlist to the database.
+     * @param p newly created Playlist object to be saved to the database
+     */
+    public void savePlaylist(Playlist p);
+
+    /**
+     * Finds a playlist by unique identifier
+     * @param id the unique identifier of the playlist to be retrieved
+     * @return Playlist with matching id or null
+     */
+    public Playlist getPlaylist(int id);
+
+    /**
+     * A method to check whether a playlist exists by id. Necessary when creating new playlists,
+     * must generate a new id and make sure that it is not already in use.
+     * @param id the integer id of the playlist object.
+     * @return true if and only if there exists a Playlist of the name submitted
+     */
+    public boolean exists(int id);
 }
--- a/src/main/java/Database/playlistAccessInterface.java
+++ b/src/main/java/Database/playlistAccessInterface.java
@@ -1,7 +1,5 @@
 package Database;
 
-import Entities.Playlist;
-
 import java.util.Collection;
 
 public interface playlistAccessInterface {
@@ -9,20 +7,20 @@ public interface playlistAccessInterface {
     /**
      * @return ArrayList of all existing Playlists
      */
-    public Collection<Playlist> getPlaylists();
+    public Collection<playlistDsData> getPlaylists();
 
     /**
      * Saves a new playlist to the database.
      * @param p newly created Playlist object to be saved to the database
      */
-    public void savePlaylist(Playlist p);
+    public void savePlaylist(playlistDsData p);
 
     /**
      * Finds a playlist by unique identifier
      * @param id the unique identifier of the playlist to be retrieved
      * @return Playlist with matching id or null
      */
-    public Playlist getPlaylist(int id);
+    public playlistDsData findPlaylist(int id);
 
     /**
      * A method to check whether a playlist exists by id. Necessary when creating new playlists,
--- a/src/main/java/Database/playlistAccessInterface.java
+++ b/src/main/java/Database/playlistAccessInterface.java
@@ -5,7 +5,7 @@ import java.util.Collection;
 public interface playlistAccessInterface {
 
     /**
-     * @return ArrayList of all existing Playlists
+     * @return Collection of all existing Playlists
      */
     public Collection<playlistDsData> getPlaylists();
 
--- a/src/main/java/Database/playlistLibrary.java
+++ b/src/main/java/Database/playlistLibrary.java
@@ -1,9 +1,5 @@
 package Database;
 
-import Entities.Playlist;
-import Entities.Song;
-import Entities.User;
-
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileWriter;
@@ -18,12 +14,10 @@ import java.util.*;
  */
 public class playlistLibrary implements playlistAccessInterface {
     private static final playlistLibrary PLAYLIST_LIBRARY =
-            new playlistLibrary("./src/main/java/Database/playlists.csv", userList.getInstance());
-    Map<Integer, Playlist> playlistDatabase;
-    String filepath;
-    songLibrary library;
-    userList users;
-    List<Integer> saved;
+            new playlistLibrary("./src/main/java/Database/playlists.csv");
+    private final Map<Integer, playlistDsData> playlistDatabase;
+    private final String filepath;
+    private final List<Integer> saved;
 
     /**
      * Global static method to retrieve the single instance of the playlistLibrary.
@@ -33,14 +27,9 @@ public class playlistLibrary implements playlistAccessInterface {
         return PLAYLIST_LIBRARY;
     }
 
-    private playlistLibrary(String path, userList users){
+    private playlistLibrary(String path){
         this.filepath = path;
         saved = new ArrayList<>();
-
-        // TODO: modify once songLibrary, userList implemented
-        library = new songLibrary();
-        this.users = users;
-
         playlistDatabase = loadFile();
     }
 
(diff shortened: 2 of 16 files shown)


src/main/java/Database/playlistAccessInterface.java
```diff
@@ -1,4 +1,34 @@
...
+import Entities.Playlist;
+
+import java.util.Collection;
+
 public interface playlistAccessInterface {
+
+    /**
+     * @return ArrayList of all existing Playlists
+     */
+    public Collection<Playlist> getPlaylists();
```
sarxhlwalker, line 12: I know we're supposed to try to be as abstract as possible in case we'll be using diff data types, but are you expecting on having different collections here? feel like the javadoc should match the actual type given here.

src/main/java/Database/playlistLibrary.java
```diff
@@ -1,4 +1,181 @@
...
+import java.io.IOException;
+import java.util.*;
+
+/**
+ * Playlist database class in the Frameworks & Drivers layer that follows the singleton design pattern, and
+ * implements an interface in the Application Business Rules layer for dependency inversion purposes.
+ * Note: since this is a singleton class, the constructor is private, and it can NOT be created with the
+ * new keyword outside this class.
+ */
+public class playlistLibrary implements playlistAccessInterface {
```
sarxhlwalker, line 19: think some of this went over my head as I'm not super familiar w the database use case, but: 
- do we care about if multiple users manage a playlist? was not sure that was accounted for in buildPlaylist and AFAIK Nick had a collaborative boolean option in his PR. I guess my question is that will collaborators ever be saved in the playlists.csv file -- if this is the case, then line 83 could possibly throw an error (giving a number of users to users.getUser()). if not, are collaborators stored *only* in the playlist object, and how are we adding collaborators? (in playlist.java the entity, an add collaborator function?) Not the most relevant to our work atm but something to keep in mind possibly. that or just deal with this all later 
- Just to be clear: this singleton manages *all* playlists ever created in our program? as opposed to the collection of playlists one particular user owns/collaborates on. in which case, would we not want RegularUser.java to have a playlistList instance attribute, so that we can present this list to the View eventually on our dashboard?

sarxhlwalker (commented):
quickly read over the code here.

src/main/java/Database/playlistAccessInterface.java
rafaelgacesa: That was a change I made to the code and missed in the javadoc, thanks for catching that. I’m returning a Collection since that’s how HashMaps work by default. Converting it to an ArrayList would require ugly casting or another loop, so I thought leaving it Collection would be fine since it’s also intervals. I’ll update the javadoc along with any other suggestions I get on this pull in next commit.

src/main/java/Database/playlistLibrary.java
rafaelgacesa: - I was assuming the collaborative definition from Spotify here, in that you can have a collaborative playlist edited by many but it still has one single owner, so line 83 would always be referring to a single user. Saving collaborators separately could be something to look at though, and it can be a simple addition in this code if @NicholasMacasaet ends up implementing it. 
- I probably could’ve explained the design better but the singleton instance of playlistLibrary acts as the single source of persistence for Playlist data. What happens to the Playlist objects once they’re loaded in is up to whoever is implementing a use case - they can choose to pull straight from the database or to store Playlist objects elsewhere. All that matters is that all Playlist objects that will persist in the program are stored by the playlistLibrary instance (and therefore playlists.csv).

NicholasMacasaet: > think some of this went over my head as I'm not super familiar w the database use case, but:
> 
> * do we care about if multiple users manage a playlist? was not sure that was accounted for in buildPlaylist and AFAIK Nick had a collaborative boolean option in his PR. I guess my question is that will collaborators ever be saved in the playlists.csv file -- if this is the case, then line 83 could possibly throw an error (giving a number of users to users.getUser()). if not, are collaborators stored _only_ in the playlist object, and how are we adding collaborators? (in playlist.java the entity, an add collaborator function?) Not the most relevant to our work atm but something to keep in mind possibly. that or just deal with this all later
> * Just to be clear: this singleton manages _all_ playlists ever created in our program? as opposed to the collection of playlists one particular user owns/collaborates on. in which case, would we not want RegularUser.java to have a playlistList instance attribute, so that we can present this list to the View eventually on our dashboard?

I think we should discuss your first point more because it sounds like adding collaborators would either be its own use case or go under the umbrella of my "playlist curation" use case.

sarxhlwalker: >     * I probably could’ve explained the design better, but the singleton instance of playlistLibrary acts as the single source of **persistence** for Playlist data. What happens to the Playlist objects once they’re loaded in is up to whoever is implementing a use case - they can choose to pull straight from the database or to store Playlist objects elsewhere. All that matters is that all Playlist objects that will persist in the program are stored by the playlistLibrary instance (and therefore playlists.csv).

Sure, but then do we not want a playlistList in RegularUser (iirc we removed playlistList from abstract class User because ArtistUser does not have playlists). Not sure if RegularUser was complete -- perhaps Olina can confirm?

rafaelgacesa: RegularUser is empty in main right now, afaik. But it would be fine to have a playlist list in user, and it probably makes sense too. It’s up to whoever is implementing the stuff around that code - the singleton design only exists to prevent multiple classes from reading/writing to the persistent database.

rafaelgacesa:
I’m going to say hold off on looking into anything besides the interfaces and the tests until I make another commit - after talking to Prof about the clean architecture, I’m going to have to abstract how the database interacts with entities. Should only take half an hour or so once I get home.

rafaelgacesa:
In my most recent commit https://github.com/CSC207-2022F-UofT/course-project-team-communify/pull/20/commits/470c34600668c3c0048f2474c99d87f9506e5d06 I added two new data structures, `userDsData.java` and `playlistDsData.java`. These classes are in the application business rules layer, and act similarly to input/output data classes: they allow entity data to be passed to an outer class (the database class in this case) without any dependencies on entity classes that would violate clean architecture.

In practice, using them is as simple as creating a new object passing the entity you'd like stored inside, and then passing that object to the databases to be saved. Or, if you are requesting an object from the database, you'll just have to call `getPlaylist()` or `getUser()` on the result to extract the entity from the data structure.

@chrstinalin you'll probably have to do something similar for songs, we should discuss soon so I don't mess up any of your plans.

I wrote most of the code in this commit on my phone in class, so there are probably some typos - just lmk if you catch something.

chrstinalin (approved):

src/main/java/Database/playlistLibrary.java
```diff
@@ -1,4 +1,181 @@
...
+    private Map<Integer, Playlist> loadFile() {
+        Map<Integer, Playlist> p = new HashMap<>();
+
+        try {
+            Scanner in = new Scanner(new File(filepath));
+            while (in.hasNextLine()){
+                // assumes 5 columns, properly filled
+                String line = in.nextLine();
+                String[] data = line.split(",");
+                p.put(Integer.parseInt(data[0]), buildPlaylist(data));
```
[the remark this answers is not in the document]
rafaelgacesa: Yeah I've heard, I chose Scanner out of familiarity since it's been a while since I've used a BufferedReader and I was trying to get this done as quickly as possible. I'll keep this in mind when I come back to complete these classes once the remaining entites/databases are implemented.

NicholasMacasaet (approved):
So far there hasn't been anything I've seen that's blatantly incorrect or needs fixing. Because how important these classes are I'll continue to look over them and contact you so we can push any more necessary changes.

NicholasMacasaet (approved):

chrstinalin (approved):

sarxhlwalker (approved):
Looks good. Thanks for fixing the autograde issues!

PR #21, opened 2022-11-07
Feature 4 playlist creation

NicholasMacasaet:
updated some javadocs in playlist and createPlaylist usecase

### DIFF
--- a/src/main/java/Entities/Playlist.java
+++ b/src/main/java/Entities/Playlist.java
@@ -17,13 +17,12 @@ public class Playlist {
         this.name = name;
         this.owner = owner;
         this.isPublic = isPublic;
-        this.songList = new LinkedList<>();
+        this.songList = new LinkedList<Song>();
     }
     public Playlist(String name, User owner, boolean isPublic, Song firstSong){
         this.name = name;
         this.owner = owner;
         this.isPublic = isPublic;
-        this.songList = new LinkedList<>();
         this.songList.addFirst(firstSong);
     }
 
--- a/src/main/java/Entities/Playlist.java
+++ b/src/main/java/Entities/Playlist.java
@@ -66,7 +66,9 @@ public class Playlist {
     public ArrayList<User> getCollaborators() {
         return this.collaborators;
     }
-
+    /**
+     * @param song removes song from playlist
+    */
     public void removeSong(Song song){
         this.songList.remove(song);
     }
--- a/src/main/java/Entities/Playlist.java
+++ b/src/main/java/Entities/Playlist.java
@@ -23,6 +23,7 @@ public class Playlist {
         this.name = name;
         this.owner = owner;
         this.isPublic = isPublic;
+        this.songList =new LinkedList<Song>();
         this.songList.addFirst(firstSong);
     }
 
--- a/src/main/java/Entities/Playlist.java
+++ b/src/main/java/Entities/Playlist.java
@@ -62,6 +62,10 @@ public class Playlist {
         return this.isPublic;
     }
 
+    /**
+     *
+     * @return list of collaborators for the playlist
+     */
 
     public ArrayList<User> getCollaborators() {
         return this.collaborators;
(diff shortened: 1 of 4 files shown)


src/main/java/Entities/Playlist.java
```diff
@@ -17,13 +17,12 @@ public Playlist(String name, User owner, boolean isPublic){
...
         this.owner = owner;
         this.isPublic = isPublic;
-        this.songList = new LinkedList<>();
+        this.songList = new LinkedList<Song>();
     }
     public Playlist(String name, User owner, boolean isPublic, Song firstSong){
         this.name = name;
         this.owner = owner;
         this.isPublic = isPublic;
-        this.songList = new LinkedList<>();
```
rafaelgacesa, line 26: Autograder fails because you deleted this line - you have to instantiate the list before calling addFirst().

rafaelgacesa (changes requested):
Added one change to make autograder work. Otherwise looks good and will approve, nice documentation.

chrstinalin (commented):
Generally looks good, but I saw theres still a 'collaborators' user list. Are we still doing collaborative playlists? 

If we are, it feels like we should have an addCollaborator. Also, it looks like you can only retrieve a list of collaborators (and it isnt initialized, either)

src/main/java/UseCase/createPlaylist.java
```diff
@@ -0,0 +1,21 @@
+package UseCase;
+import Entities.Song;
+import Entities.User;
+import Entities.Playlist;
+import InputBoundary.newPlaylistInputBoundary;
+
+public class createPlaylist implements newPlaylistInputBoundary {
+
+    /**
+     @param song will only be used if creating a nonempty playlist with one song
```
manya-mittal, line 10: The use of polymorphism is good here but could you please move the javadoc to the second method? Since the second method uses the song parameter, I think it makes sense to use the javadoc to describe the use of the parameter in the second method.

manya-mittal (changes requested):

chrstinalin:
Also -- the firstSong -- I'm assuming this is the case where the user goes from a given song's dropdown menu directly to new playlist? 

If I recall correctly, there's also the possibility of creating an empty playlist (via the playlist menu directly). May want to accommodate that?

src/main/java/UseCase/createPlaylist.java
NicholasMacasaet: > Generally looks good, but I saw theres still a 'collaborators' user list. Are we still doing collaborative playlists?
> 
> If we are, it feels like we should have an addCollaborator. Also, it looks like you can only retrieve a list of collaborators (and it isnt initialized, either)

For now I think it would be best to leave it, if we don't decide to do I can delete the collaborator option in the next push

manya-mittal (dismissed):
Thanks for making the changes I requested Nick! This looks great now.

rafaelgacesa (dismissed):
Everything looks good now.

chrstinalin (dismissed):
Looks good 👍

src/main/java/InputData/newPlaylistInputData.java
```diff
@@ -0,0 +1,47 @@
+package InputData;
+import Entities.Song;
+import Entities.User;
+import java.util.Optional;
+
+public class newPlaylistInputData {
+    private final String playlistName;
+    private final User owner;
+    private final Song firstSong;
+    private boolean privacy;
```
manya-mittal, line 10: Could we rename this to be isPrivate? I think that would be a better name since it's a boolean variable and it would also be more consistent with the naming convention you used in your other classes (since you previously used isPublic)

NicholasMacasaet: Ok, I will update that.

manya-mittal (approved):
This looks great now!

rafaelgacesa (approved):
She really made you dismiss approvals over a variable name 💀

chrstinalin (approved):
💀💀💀

PR #22, opened 2022-11-08
Replaced String filePath with File file, skeletal stuff

chrstinalin:
**Song Entity Change**

Original design would have forced playSong/playSpace to directly access the database in order to play the mp3, violating clean architecture.

A better alternative might be a Media object if the only purpose the File serves is to be played, but 1) I don't know if anyone else needs to access the file for a different purpose, and 2) I'm not certain which audio-playing library @Mistryr222 @sarxhlwalker  you guys plan on using. From what I can tell, the common method (which uses Media) requires javafx to be installed. That might be something you guys want to come to an agreement on.

I'm committing this as a File for now just to ensure nobody references the original string fileLocation.

**Skeleton Stuff**

Just put some rough skeletal stuff down, figured I might as well upload it too. Alot of stuff is regurgitated from the CRC Model doc. Also added uploadSongOutputData, but generally didn't really focus on this part.

Main focus was on restructuring things to include a songDsData in light of the clean architecture violation @rafaelgacesa pointed out [here](https://github.com/CSC207-2022F-UofT/course-project-team-communify/pull/20#issuecomment-1302574894). Let me know if I misinterpreted anything.

### DIFF
--- a/src/main/java/Controller/uploadSongController.java
+++ b/src/main/java/Controller/uploadSongController.java
@@ -1,4 +1,11 @@
 package Controller;
+import InputBoundary.uploadSongInputBoundary;
 
 public class uploadSongController {
+
+    uploadSongInputBoundary uploadSongInteractor;
+    public void upload(uploadSongInputBoundary uploadSongInteractor){
+        //TODO
+    }
+
 }
--- a/src/main/java/Database/songAccessInterface.java
+++ b/src/main/java/Database/songAccessInterface.java
@@ -1,4 +1,27 @@
 package Database;
+import java.util.Collection;
 
 public interface songAccessInterface {
+    /**
+     * @return Collection of all songs.
+     */
+    public Collection<songDsData> getLibrary();
+
+    /**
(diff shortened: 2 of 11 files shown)


sarxhlwalker (approved):
I think passing around the mp3 file makes sense and is cleaner than returning to the database to play the song. As discussed on Discord, I'm not gonna be playing the song itself but rather calling Rohan's playSong method, so he can pick what method he wants to play the song.

rafaelgacesa (approved):
Agree with the change to the Song class, a File object should be functionally equivalent.

Use case skeletons are fine, pretty much as according to our original deisgn.

The songDsData class is done as I imagined, no misinterpretation. The basic implementation of the database is good, nice implementation of Singleton too.

sarxhlwalker (approved):

manya-mittal (approved):
I agree with the changes you made. Passing around the actual file will be easier than trying to access the file through the file path. The skeleton classes also look great!

PR #23, opened 2022-11-09
Feature 2 registration

SiqiZhang6:
made the ArtistUser, ArtistUserFactory, and userFactory entities. For regular users implementing userFactory they would pass in a empty string for artistname. please let me know if that works or if it needs to be changed.


rafaelgacesa:
I think we're going to have to re-think this Factory design - having the factory interface `create()` method with all three parameters violates the interface segregation principle, since the userFactory would then have the artist name parameter it does not need. I think we may need to scrap the interface, but I need to brush up on the Factory design pattern. Some other perspectives could be helpful? @chrstinalin @Mistryr222 @sarxhlwalker

src/main/java/Entities/userFactory.java
```diff
@@ -1,4 +1,5 @@
 package Entities;
 
 public interface userFactory {
+    User createUser(String artistName, String username, String password);
```
sarxhlwalker, line 4: Re @rafaelgacesa's comment about interface segregation principle: technically, it's fine because that specifies reliance on a _method_ that is not used (to create a regular user, I assume the userFactory will still be called; it's the parameter artistName that is unused). 

Instead, I would recommend overloading the method createUser in the interface -- i.e. have User createUser(artistName, username, password) as you do now and add another line underneath with User createUser(username, password) so that when you're creating a regular user, you don't need to pass anything in for the artist name at all.

sarxhlwalker (changes requested):
See comment.

src/main/java/Entities/userFactory.java
rafaelgacesa: Wouldn’t overloading it then create a violation of the interface segregation principle ?

chrstinalin:
The easy way out would be to give regularUsers a name as well (as it's not necessarily a bad thing to have) and generalize the artistName in User to just name. I don't know if that bothers anyone, but I don't know if something like this is necessarily worth the struggle (unless someone wants to use it as an opportunity to learn).

src/main/java/Entities/userFactory.java
sarxhlwalker: Hmm... maybe you're right. I didn't think you'd have to implement both versions when implementing the interface, but now that I think about it, you probably do. In that case, what Jessica currently has should be fine?

sarxhlwalker:
For some reason, GitHub won't let me reply, but I think that's fine. I'm not sure when we'll ever use the regular user's name though? Like, why do we ask for that information if we never use it aside from making our code easier?

chrstinalin:
The easy way out would be to give regularUsers a name as well (as it's not necessarily a bad thing to have) and generalize the artistName in User to just name. I don't know if that bothers anyone, but I don't know if something like this is necessarily worth the struggle (unless someone wants to use it as an opportunity to learn). Putting it out there as a last resort.

sarxhlwalker:
> The easy way out would be to give regularUsers a name as well (as it's not necessarily a bad thing to have) and generalize the artistName in User to just name. I don't know if that bothers anyone, but I don't know if something like this is necessarily worth the struggle (unless someone wants to use it as an opportunity to learn). Putting it out there as a last resort.
> 
> (sorry Sarah you read my message too fast -- I deleted the original because of a spelling error, lol)
> 
> I figure we could use it in a welcome message on the dashboard, but that's about the extent of it.

OK yeah, that would be fine IMO

src/main/java/Entities/userFactory.java
sarxhlwalker: @chrstinalin  do you mean my initial comment or raf's suggestion? I realize now that my initial suggestion wouldn't really help lol.

chrstinalin: I was replying to your original comment since that's the only thing that was coming up for me, lol. I agree with what he's saying.

chrstinalin (changes requested):
See [here](https://github.com/CSC207-2022F-UofT/course-project-team-communify/pull/23#issuecomment-1309207525).

src/main/java/Entities/userFactory.java
SiqiZhang6: I went to the office hour and the ta suggested to just have the userFactory for creating both user object, what do you guys think?

chrstinalin: One user factory with two methods?

PR #25, opened 2022-11-09
Feature 9 listen to space

sarxhlwalker:
While I'm here, some basic changes. 
- created the Main.java file -- I've been playing around with Swing in order to get the Views/ViewModels/Presenter working, and to have anything show up, we will seem to need a Main.java file. Will push actual code in there once it works. 
- Basic use case for listen to space, calling @Mistryr222's use case -- I was under the impression I'd pass to you the song file itself, but let me know if you'd rather I pass the songInputData as a whole. (There is a dummy version of that file attached.) 
- Just wanted to confirm with everyone that we will _not_ be outputting any kind of confirmation to the user that a song has been added to a space, as we have no OutputBoundary/etc. for this case. (IIRC we decided that there would be a popup for add to playlist, so am wondering if we want consistency.)

### DIFF
--- a/src/main/java/Entities/Song.java
+++ b/src/main/java/Entities/Song.java
@@ -1,6 +1,7 @@
 package Entities;
 
 import java.awt.image.BufferedImage;
+import java.io.File;
 import java.util.List;
 
 public class Song {
@@ -10,7 +11,7 @@ public class Song {
     private final List artistList;
     private final int length;
     private final String genre;
-    private final String fileLocation;
+    private final File file;
     private final boolean isExplicit;
     private final BufferedImage cover;
 
@@ -21,18 +22,18 @@ public class Song {
      * @param artistList List of contributing artists.
      * @param length Length of the song, in seconds.
      * @param genre Type of song.
-     * @param fileLocation Directory path to audio file.
+     * @param file Audio file itself.
      * @param isExplicit Whether the song is explicit or not.
      * @param cover Cover image of the song.
      */
     public Song(int id, String name, List artistList, int length, String genre,
-                String fileLocation, boolean isExplicit, BufferedImage cover){
+                File file, boolean isExplicit, BufferedImage cover){
         this.id = id;
         this.name = name;
         this.artistList = artistList;
         this.length = length;
         this.genre = genre;
-        this.fileLocation = fileLocation;
+        this.file = file;
         this.isExplicit = isExplicit;
         this.cover = cover;
     }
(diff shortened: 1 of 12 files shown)


src/main/java/UseCase/playSpaceInteractor.java
```diff
@@ -1,16 +1,34 @@
...
-import OutputBoundary.spacePlayedOutputBoundary;
 import OutputData.songOutputData;
+import java.io.File;
 
+/**
+ * use case for playing a space
+ */
 public class playSpaceInteractor {
 
     private OutputBoundary.spacePlayedOutputBoundary spacePlayedOutputBoundary;
```
manya-mittal, line 13: You would need to have a constructor to initialise this instance variable I think. This would allow us to pass in a Presenter object so that when you call the spacePlayed method later on, you are actually calling it from the Presenter class.

manya-mittal (changes requested):
Other than the constructor, everything else seems good. I'll approve the pull request after the constructor is added :)

manya-mittal (approved):
These changes look great! I agree that plassing the File object is easier than having to pass the file path and then look up the file. The javadocs also all look good and will be useful to refer to in the future.

src/main/java/Presenter/spacePresenter.java
```diff
@@ -6,14 +6,24 @@
 
 import java.util.ArrayList;
 
+/**
+ * presenter for the space use cases.
+ */
 public class spacePresenter implements songOutputBoundary, spacePlayedOutputBoundary {
```
SiqiZhang6, line 12: remember to override the methods in songOB and spacePlayedOB. (ie, change the button and the playbar). if they are the same then merge

SiqiZhang6 (commented):
at spacePresenter.java line 12
remember to override the methods in songOB and spacePlayedOB. (ie, change the button and the playbar). if they are the same then merge

src/main/java/Entities/Song.java
```diff
@@ -10,7 +11,7 @@ public class Song {
     private final List artistList;
     private final int length;
     private final String genre;
-    private final String fileLocation;
+    private final File file;
```
Mistryr222, line 14: I know there was some debate about keeping the song file as an instance of the File class or just a String path, but keep it as a File instance because it makes it easier for testing and makes more sense to pass a file around rather than a string.

Mistryr222 (changes requested):

Mistryr222 (approved):
The changes look good!

manya-mittal (approved):
This looks great to me!

SiqiZhang6 (approved):
the changes look great :D

PR #26, opened 2022-11-09
Feature 2 register

SiqiZhang6:
updated ArtistUser and combined the factories into one factory.


chrstinalin (approved):
Looks good to me

rafaelgacesa (approved):
This is still technically not a classic Factory design, but it looks like it'll work fine.

sarxhlwalker (approved):
Looks good to me.

PR #27, opened 2022-11-10
Feature 6 search engine

manya-mittal:
I made some significant progress for my use case. Here is a summary: 

- Coded the interfaces i.e searchInputBoundary and searchOutputBoundary
- Coded the Search class (the Use Case Interactor). This is the most significant class in this pull request because it has the search method, which is the one that does the actual work. 
- Coded a few other basic classes, including the Controller, output data and inputdata 
- Added some javadocs. I wasn't sure if javadocs were needed for constructors and getters/setters so I left those out for now.

### DIFF
--- a/src/main/java/Controller/searchController.java
+++ b/src/main/java/Controller/searchController.java
@@ -1,4 +1,18 @@
 package Controller;
 
+import InputBoundary.searchInputBoundary;
+import InputData.searchInputData;
+
 public class searchController {
+    private final searchInputBoundary searchInteractor;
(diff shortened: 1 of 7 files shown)


rafaelgacesa (approved):
Nothing in this pull offends me. Looks like some good progress on your use case! 🫡

sarxhlwalker (approved):
Looks good to me!

SiqiZhang6 (approved):
looks great! nice work :D

PR #29, opened 2022-11-11
Feature 7 Audio System - playAudio() initial pull

Mistryr222:
There are still many changes that need to be added as written in each of my Java docs but I want to pull now so if Sarah wants to make test cases she can.

1. Made small fixes to other peoples files, super simple ones, but still important, example: adding a missing this. keyword
2. Added the playAudio(File songFile) use case, to be used by my playSong() use case and Sarah's playspace
3.  Still need to add a pause audio feature and make the "Player" follow the singleton principle.


src/main/java/UseCase/playAudioInteractor.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Add Java docs later, THIS METHOD NEEDS TO CHANGE AFTER PLAYER IS MADE INTO SINGLETON
+     * HAS SOME EXCEPTIONS THAT NEED TO BE HANDLED AS WELL
+     * @param songFile
+     */
+    @Override
+    public void playAudio(audioInputData songFile) throws FileNotFoundException, JavaLayerException {
+
+        FileInputStream media = new FileInputStream(songFile.getSongFile().getAbsolutePath());
```
rafaelgacesa, line 33: I'll wait for other comments/reviews before I review this, but depending on what happens with my pull this may all be unnecessary.

rafaelgacesa (commented):
Waiting for discussion on the other active pull related to playing songs before making a decision.

src/main/java/InputData/editPlaylistInputData.java
```diff
@@ -28,7 +28,7 @@ public Song getSong(){
         return this.song;
     }
     public String getNewName(){
-        return newName;
+        return this.newName;
```
NicholasMacasaet, line 31: Thanks for catching this part in my editPlaylist and newPlaylist input data classes, I missed this

src/main/java/UseCase/playSong.java
```diff
@@ -10,3 +10,5 @@ static void playAudio(File songFilePath){
         // TODO: call presenter
```
sarxhlwalker, line 10: i just realized that on line 7 of this file (above your changes) it should be playAudio(File songFile) as opposed to playAudio(File songFilePath). that was probably a carry-over from our discussion about switching back to songFilePath, my bad.

src/main/java/Controller/playAudioController.java
```diff
@@ -0,0 +1,29 @@
...
+
+    private final playAudioInputBoundary playAudioInteractor;
+    private final audioInputData songFile;
+
+    /**
+     * Add Java docs after
+     */
+    public playAudioController(playAudioInputBoundary playAudioInteractor, audioInputData songFile) {
+        this.playAudioInteractor = playAudioInteractor;
+        this.songFile = songFile;
```
sarxhlwalker, line 19: I might be wrong, but I was under the impression that the controller _constructs_ the audioInputData as opposed to simply being passed the already-existing audioInputData (the alternative would be the viewModel constructing the audioInputData, which IIRC we discussed in office hours and was incorrect/violates clean architecture). Could someone confirm? (I bring this up because I was wondering about how to approach this in my own use case.)

sarxhlwalker (commented):
Comments -- also waiting for decision on Raf's PR -- also, I think the GitHub auto-grade on this PR fails.

src/main/java/UseCase/playSong.java
rafaelgacesa, line 10: This class needs to extend the input boundary.

PR #30, opened 2022-11-11
Created the music engine

rafaelgacesa:
This is some of the most difficult to follow code that I have written, so I apologize in advance.
This will definitely need to be a topic of discussion, and maybe even something to bring to office hours.

### In this pull:
- The `JPlayer` entity
   - This is a modified version of `Player` and `AdvancedPlayer` from `JLayer` I had to put together to better suit our needs
   - I left a note at the top, it basically says the code mostly isn't ours and it shouldn't be marked; pretend that it's dependency code
- The singleton `MusicPlayer` entity
   - This entity handles actually playing and pausing music
      - It's singleton for pretty good reason - having more than one object playing audio would be messy
   - The controversy is that there is **essentially use-case code inside of an entity**
      - This solves all of our problems but I have no clue how bad of a violation it is

Let me know what you think.

### DIFF
--- a/src/main/java/Entities/JPlayer.java
+++ b/src/main/java/Entities/JPlayer.java
@@ -1,0 +1,266 @@
+/**
+ * Note to TAs: This code is from a dependency and is only
+ * slightly modified to suit our needs better. Please ignore this file and assume that it is part
+ * of a package we used - we did not write it, and we do not intend for it to be marked.
+ */
+
+package Entities;
+
+/*
+ * 11/19/04		1.0 moved to LGPL.
+ * 29/01/00		Initial version. mdm@techie.com
+ *-----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Mistryr222 (approved):
We need to remember to speak to the TA and maybe a prof about using the code which is obviously fine, but just to be even more safe. An entity having "use case code" can probably be fixed - another thing for an office hour. Otherwise, looks good.

sarxhlwalker (approved):
I like the idea of the MusicPlayer being a singleton, I think that makes sense. I agree that we should go to office hours about it being a "use case entity", though, and confirm that this method follows clean architecture.

rafaelgacesa:
Re-opening this pull in a different branch where it makes more sense, with new updates to the classes.

PR #33, opened 2022-11-11
Updated audio engine

rafaelgacesa:
Cleaned up the code and added synchronization. Most of this pull is similar to the original, just in a new branch.
This will definitely need to be a topic of discussion, and maybe even something to bring to office hours.
Ignore all of the other commits in this pull, I didn't include the code in them in this push.

### In this pull:
- The `JPlayer` entity
   - This is a modified version of `Player` and `AdvancedPlayer` from `JLayer` I had to put together to better suit our needs
   - I left a note at the top, it basically says the code mostly isn't ours and it shouldn't be marked; pretend that it's dependency code
- The singleton `MusicPlayer` entity
   - This entity handles actually playing and pausing music
      - It's singleton for pretty good reason - having more than one object playing audio would be messy
   - Now include synchronization using threads for queuing things
   - The controversy is that there is **essentially use-case code inside of an entity**
      - This solves all of our problems but I have no clue how bad of a violation it is

Let me know what you think.

### DIFF
--- a/src/main/java/Entities/JPlayer.java
+++ b/src/main/java/Entities/JPlayer.java
@@ -1,0 +1,266 @@
+/**
+ * Note to TAs: This code is from a dependency and is only
+ * slightly modified to suit our needs better. Please ignore this file and assume that it is part
+ * of a package we used - we did not write it, and we do not intend for it to be marked.
+ */
+
+package Entities;
+
+/*
+ * 11/19/04		1.0 moved to LGPL.
+ * 29/01/00		Initial version. mdm@techie.com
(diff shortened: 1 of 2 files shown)


Mistryr222 (approved):
Updated as requested. Looks good.

chrstinalin (approved):
Makes sense to me, though we should probably double-check at an office hour to be safe.

sarxhlwalker (approved):
As before -- a singleton for the music player makes sense, but I agree we should double check with a prof in office hours.

PR #37, opened 2022-11-12
Completed databases, fixed user entities, and other small changes

rafaelgacesa:
General completion of playlist and user databases and database-related entity classes.

### In this pull:

**Entities:**
- finished the `RegularUser` class with a `playlistList` and everything else required
- added some constructors and a `songs` attribute to `artistUser`
- added corresponding required methods to `userFactory`

**Database related classes:**
- updated `playlistLibrary` to save properly so playlists can be updated
   - will be useful for @NicholasMacasaet
   - also did a general refactor of the file I/O methods to simplify them
   - updated `playlistDsData` to deal with privacy changes
- completed the `userList` database since the entities are now complete
   - pretty straightforward, created tests for it
   - updated `userDsData` to create the `User` objects and build their `.csv` output in a `toString()` method
- added some small changes to @chrstinalin 's `songLibrary` to allow tests to pass
   - these changes are all pretty temporary, marked with `TODO`s
   - some changes also made to `songDsData` in a similar vein

### DIFF
--- a/src/main/java/Database/playlistDsData.java
+++ b/src/main/java/Database/playlistDsData.java
@@ -13,6 +13,7 @@ import java.util.LinkedList;
  */
 public class playlistDsData {
     private final Playlist playlist;
+    private final int privacy;
     private final songAccessInterface library;
     private final userAccessInterface users;
     
@@ -21,12 +22,14 @@ public class playlistDsData {
         this.playlist = p;
         library = songLibrary.getInstance();
         users = userList.getInstance();
+        this.privacy = 1;
     }
     
     public playlistDsData(String[] data){
         library = songLibrary.getInstance();
         users = userList.getInstance();
         this.playlist = buildPlaylist(data);
+        this.privacy = Integer.parseInt(data[4]);
     }
 
     /**
@@ -41,14 +44,17 @@ public class playlistDsData {
 
         String[] songs = data[2].split(";");
         User owner = users.getUser(data[3]).getUser();
-        boolean isPublic = Boolean.parseBoolean(data[4]);
         Playlist playlist = new Playlist(id, name, owner);
 
         for (String song : songs){
             if (song.length() > 0) {
+                playlist.addSong(library.getSong(Integer.parseInt(song)).getSong());
+                // TODO: add this check back in once songLibrary is complete
+                /*
                 if (library.exists(Integer.parseInt(song))) {
                     playlist.addSong(library.getSong(Integer.parseInt(song)).getSong());
                 }
+                 */
             }
         }
 
@@ -68,7 +74,7 @@ public class playlistDsData {
             songIds.add(songID);
         }
         return id + "," + playlist.getName() + "," + String.join(";", songIds) + "," +
-                playlist.getOwner().getUsername();
+                playlist.getOwner().getUsername() + "," + privacy;
     }
 
     /**
--- a/src/main/java/Database/playlistLibrary.java
+++ b/src/main/java/Database/playlistLibrary.java
@@ -45,8 +45,10 @@ public class playlistLibrary implements playlistAccessInterface {
                 // assumes 5 columns, properly filled
                 String line = in.nextLine();
                 String[] data = line.split(",");
-                int id = Integer.parseInt(data[0]);
-                p.put(Integer.parseInt(data[0]), new playlistDsData(data));
+                if (data.length == 5) {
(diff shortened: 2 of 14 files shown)


NicholasMacasaet (approved):
Everything looks straightforward enough. The implementation for the RegularUser things that are necessary for my use case look great. If I have any questions about the database that are relevant to my code I will make sure to ask you. But other than that good work Raf.

src/main/java/Database/playlistLibrary.java
```diff
@@ -77,13 +77,11 @@ private void createFile(String fp){
      */
```
NicholasMacasaet, line 77: Small change, maybe add some Javadocs here/descriptions

NicholasMacasaet (changes requested):
Small change, maybe add some javadocs for the saveFile method

src/main/java/Database/playlistLibrary.java
rafaelgacesa: I can't tell which method you're highlighting here. Unless you're asking for some more comments within the methods on logic (which I definitely should add eventually lol), I don't think there's any methods missing javadocs?

NicholasMacasaet: Ah sorry this is the saveFile() method, but just adding a small description would be good.

NicholasMacasaet (approved):

src/main/java/Database/songLibrary.java
```diff
@@ -9,9 +9,10 @@
  */
 public class songLibrary implements songAccessInterface{
 
+    // TODO: change "" param to filepath
     private static final songLibrary SONG_LIBRARY = new songLibrary("");
     private final HashMap<Integer, songDsData> library;
-    private String filepath;
+    private final String filepath;
```
chrstinalin, line 15: Whoops

src/main/java/Database/songLibrary.java
```diff
@@ -9,9 +9,10 @@
  */
 public class songLibrary implements songAccessInterface{
 
+    // TODO: change "" param to filepath
```
chrstinalin, line 12: Ah yeah my local file's had it changed for a while. I still need to make a pull, but I wanted to finish things up.

chrstinalin (approved):
Looks good!

src/main/java/Database/songLibrary.java
rafaelgacesa: Lol, I was just clicking on all the random warnings when I pulled.

src/main/java/Database/playlistDsData.java
```diff
@@ -21,12 +22,14 @@ public playlistDsData(Playlist p){
         this.playlist = p;
         library = songLibrary.getInstance();
         users = userList.getInstance();
+        this.privacy = 1;
```
sarxhlwalker, line 25: I'd suggest adding a comment here to signify what 1 (vs. 0 presumably) signifies wrt privacy (also I thought we were getting rid of playlist privacy controls? is that still the case?

src/main/java/Database/playlistLibrary.java
```diff
@@ -46,9 +45,10 @@ private Map<Integer, playlistDsData> loadFile() {
                 // assumes 5 columns, properly filled
                 String line = in.nextLine();
                 String[] data = line.split(",");
-                int id = Integer.parseInt(data[0]);
-                saved.add(id);
-                p.put(Integer.parseInt(data[0]), new playlistDsData(data));
+                if (data.length == 5) {
+                    int id = Integer.parseInt(data[0]);
+                    p.put(id, new playlistDsData(data));
+                }
```
sarxhlwalker, line 51: Should there be some kind of warning of result if the 5 columns are not properly filled and you don't complete loading the file?

sarxhlwalker (approved):
Overall looks good. Some comments.

src/main/java/Database/playlistLibrary.java
rafaelgacesa: It just skips the line without 5 columns and moves on, so it should be fine. Theoretically, my database file is the only thing that should be touching the csv files so I think it's an ok assumption.

src/main/java/Database/playlistDsData.java
rafaelgacesa: Yeah, that is gone from the use cases but I don't want to waste time removing it from the databases until we're totally sure it needs to go.

PR #39, opened 2022-11-12
space stuff

sarxhlwalker:
A professor in office hours suggested that we remove the Space entity and instead store the collection of songs to play in the spaceController. This PR reflects this change. 

Some notes: 
- I never ended up using spaceAddInputBoundary -- @SiqiZhang6, do you think this is a file OK to remove? (I did not delete it in this PR.) 
- the RegularUser is the same as the one in Raf's PR #37. 
- a (rather ugly) version of the playlistView is initialized, because I figure people will want to start adding to it soon. 
- the call to playSpace() within playlistView is commented out because it throws an error since the singleton songLibrary is not initialized and therefore can't pick a random song from it. I'm not 100% sure where initializing this class would take place -- Main.java perhaps? @chrstinalin do you have any thoughts? 
- a dummy version of songInputData has been implemented. 

Off the top of my head, what's left for spaces: 
- interacting with @manya-mittal's view to finalize adding a song to the space (i.e. clicking "add to space", implementing the chain of function calls that will eventually add a song to the collection within spaceController, returning a pop-up) 
- creating and updating the playbar on the playlistView after playing a space. 
- reverting the space button to its default state when no longer playing a space. 
- unit tests.


rafaelgacesa:
The `songLibrary` is not properly initialized as the class is not finished yet. Once @chrstinalin finishes the class, it initializes itself (according to singleton design). If you want the tests to pass, I have a slightly modified version of `songLibrary` in most of my recent pulls that works.

src/main/java/InputData/songInputData.java
```diff
@@ -1,4 +1,23 @@
 package InputData;
 
+import Entities.Song;
+
 public class songInputData {
```
rafaelgacesa, line 5: lol this is identical to my dummy version

rafaelgacesa (dismissed):
Looks good! No glaring problems. Thanks for taking initiative on the view, very helpful.

src/main/java/Controller/spaceController.java
```diff
@@ -1,41 +1,100 @@
...
      */
-    public void spaceAddSong(spaceInputData spaceInputData){
-        this.spaceAddInteractor.spaceAddSong(spaceInputData);
+    public void spaceAddSong(songInputData songInputData){
+        Integer songToAddID = songInputData.getSong().getID();
+        for (songInputData currSongInputData : this.spaceSongList){
+            Integer currSongID = currSongInputData.getSong().getID();
+            if (currSongID.equals(songToAddID)){
+                return;  // if the song is already in the playlist, do nothing
+                // TODO: make this cooler (i.e. upvote algo) if time
```
chrstinalin, line 92: If you do implement the upvote algorithm, you might want to keep in mind it wouldn't _just_ be that -- you'd have to have the GUI that comes with it, which would either 1) significantly affect the layout of the dashboard or 2) give space it's own separate page that shows everything in the user-requested queue. (likely the latter). 

tldr this is less of a 'if we have time' feature and more of a 'plan in advance' one

chrstinalin (approved):
Looks good! Kind of excited to see this play out, lol. (no pun intended)

No need to worry about initializing the library, as Raf said. If you're curious as to why, you can take a look at the comment on top of songLibrary.java. There's a brief explanation as to why (Or, well -- tl;dr, when JVM first loads the file, it'll call the static initializer SONG_LIBRARY and in hand create the object).

Also -- once I figure out why I'm having jlayer and jaudiotagger import issues, finishing up songLibrary should be incredibly quick. I'll let you guys know.

src/main/java/Controller/spaceController.java
sarxhlwalker: I was initially thinking of instead having the upvote process itself on the GUI, it'd just be handled within that function -- i.e. make it so that when you add a song to the space that is already queued, it might update the song's position within the queue, but there's no explicit "upvote" button. I could conceivably see changing the output to the presenter (if indeed we're doing that -- last I remember, add to space doesn't have an output data, @SiqiZhang6 could you confirm?) to say something like "song recommended to be played sooner!" instead of "song added to space!" or something, but nothing major.

chrstinalin: >make it so that when you add a song to the space that is already queued, it might update the song's position within the queue, but there's no explicit "upvote" button. 
If it's just that, I don't see a reason why we couldn't just do it now, lol.

sarxhlwalker: I mean we probably could, I just haven't figured out what that will look like in actual code, lol. Like what determines an increase in 1 position? At what point do you stop increasing (assuming we want the thing where we weigh both time in the queue + popularity)? etc. Once that's settled, I think actually implementing it shouldn't be that bad.

SiqiZhang6: > last I remember, add to space doesn't have an output data, @SiqiZhang6 could you confirm?

yes there is no output data, unless you wanna add one

SiqiZhang6 (approved):
Yes I agree that we can remove spaceAddInputBoundary  since we're using the song input instead.  Nice work!

chrstinalin (approved):

SiqiZhang6 (approved):

rafaelgacesa (approved):

PR #40, opened 2022-11-12
Controller class for create playlist added & modification of use case done

NicholasMacasaet:
Modified use case to save the newly created playlist to the User's playlist list & created controller class for this use case. Although, unsure if for the Controller, if this is the layer that needs to send a string notification to the presenter so I would like some feedback on this


rafaelgacesa (dismissed):
The casting is a bit awkward and something to maybe discuss at an office hour but everything else seems fine at the moment.

src/main/java/Controller/newPlaylistController.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     *
+     * @param id Assign a unique ID to this playlist
+     * @param owner Assign an Owner to this playlist
+     * @param playlistName Assign a name to this playlist
+     * @param firstSong Create a new playlist with a singular song
+     * @return String confirmation of the playlist's creation and addition to playlist database
+     */
+    public String createPlaylist(int id, User owner, String playlistName, Song firstSong){
```
sarxhlwalker, line 36: If you make the owner parametera  RegularUser (I believe artistUsers _cannot_ own playlists), then that should remove the need for casting in createPlaylist.java, no?

sarxhlwalker (commented):
See comment.

src/main/java/Controller/newPlaylistController.java
NicholasMacasaet: That is true but im also unsure whether or not the parameters for this should be the id, owner and name or the input data

NicholasMacasaet: I looked at your implementation for the controller and I saw that

NicholasMacasaet:
The view model is most definitely incorrect, but if possible I'd like feedback about whether my modification of the use case to pass the playlist notification message as an output data object is correct

src/main/java/ViewModel/playlistViewModel.java
```diff
@@ -1,4 +1,24 @@
...
+
+import Presenter.playlistPresenter;
 public class playlistViewModel {
+    private final newPlaylistController newPlaylistController;
+    private final playlistPresenter playlistPresenter;
+
+    public playlistViewModel(){
+        this.playlistPresenter = new playlistPresenter();
+        newPlaylistInputBoundary newPlaylistInputBoundary = new createPlaylist();
+        this.newPlaylistController = new newPlaylistController(newPlaylistInputBoundary, this.playlistPresenter);
```
sarxhlwalker, line 16: This requires an input data variable, based on the way you wrote newPlaylistController.java. In actuality though, I believe the controller is supposed to create the input data, not get it from the ViewModel. I believe this is why your autograde fails.

sarxhlwalker (changes requested):

src/main/java/ViewModel/playlistViewModel.java
sarxhlwalker: I don't think you need a newPlaylistInputBoundary either, because IBs are supposed to wrap controllers -- why would you pass a controller to the construction of a new controller?

src/main/java/ViewModel/playlistViewModel.java
```diff
@@ -1,4 +1,37 @@
...
+    public playlistViewModel(int id, RegularUser user, String playlistName, Song firstSong){
+        // get necessary information from View to construct a new playlist
+        newPlaylistInputData inputData = new newPlaylistInputData(id,playlistName,firstSong,user);
+        //Blank output data object to extract success message from
+        newPlaylistOutputData outputData = new newPlaylistOutputData("");
+        this.playlistPresenter = new playlistPresenter(this, outputData);
+//        newPlaylistInputBoundary newPlaylistInputBoundary = InputBoundary.newPlaylistInputBoundary;
+        createPlaylist newPlay= new createPlaylist();
+        this.newPlaylistController = new newPlaylistController(newPlay,inputData);
+        //TODO: Does this violate clean architecture ?
```
NicholasMacasaet, line 24: Not entirely sure if this is legal but I implemented the presenter/view stuff according to how Sarah described

src/main/java/Controller/newPlaylistController.java
```diff
@@ -0,0 +1,28 @@
...
+import InputBoundary.newPlaylistInputBoundary;
+import InputData.newPlaylistInputData;
+
+/**
+ * Controller-layer implementation for create playlist use case
+ */
+public class newPlaylistController {
+    private final newPlaylistInputBoundary newPlaylistInputBoundary;
+    private final newPlaylistInputData newPlaylistInputData;
+    public newPlaylistController(newPlaylistInputBoundary newPlaylistInputBoundary,newPlaylistInputData newPlaylistInputData){
```
sarxhlwalker, line 11: Again, I was under the impression that the controller should be constructing the use case interactor instead of being given it from the viewModel. if this is incorrect, please let me know.

src/main/java/InputData/newPlaylistInputData.java
```diff
@@ -1,22 +1,23 @@
 package InputData;
+import Entities.RegularUser;
 import Entities.Song;
 import Entities.User;
 
 public class newPlaylistInputData {
     private final int id;
     private final String playlistName;
-    private final User owner;
+    private final RegularUser owner;
```
sarxhlwalker, line 9: If you want, you could add a comment explaining why the owner is a RegularOwner (instead of User) is a valid assumption to make.

src/main/java/Presenter/playlistPresenter.java
```diff
@@ -1,4 +1,30 @@
...
+import OutputBoundary.newPlaylistOutputBoundary;
+import OutputData.newPlaylistOutputData;
+import ViewModel.playlistViewModel;
 
-public class playlistPresenter {
+public class playlistPresenter implements newPlaylistOutputBoundary {
+
+    String outputMessage;
+
+    final playlistViewModel viewModel;
```
sarxhlwalker, line 10: I assume this is not meant to be default and instead private?

src/main/java/Presenter/playlistPresenter.java
```diff
@@ -1,4 +1,30 @@
...
+    }
+    //TODO implement later & ask for help on. Also add editPlaylist stuff when I have better handle on output stuff
+
+    /**
+     *
+     * @param newPlaylistOutputData message to confirm new creation of playlist
+     * @return String confirming the creation of playlist generated by the createPlaylist Use case
+     */
+    public String playlistCreationConfirmation(newPlaylistOutputData newPlaylistOutputData) {
+        return viewModel.callNewPlaylistUseCase();
```
sarxhlwalker, line 28: Why are you passing newPlaylistOutputData into the function and then not using it? Presumably viewModel.callNewPlaylistUseCase() should take newPlaylistOutputData as an input, then.

src/main/java/ViewModel/playlistViewModel.java
```diff
@@ -1,4 +1,37 @@
...
+    }
+
+    /**
+     *
+     * @return output message generated by the Controller and update the presenter's outputMessage with it as well
+     */
+    public String callNewPlaylistUseCase(){
+        String msg =  this.newPlaylistController.createPlaylist();
+        newPlaylistOutputData outputData = new newPlaylistOutputData(msg);
```
sarxhlwalker, line 33: I'm a bit confused as to what you're trying to do in this function. Could you add comments? 

Also, see my comment on playlistPresenter -- if you pass the string message in from the Presenter as a parameter, you can just return it immediately.

src/main/java/UseCase/createPlaylist.java
```diff
@@ -5,26 +5,41 @@
 import InputData.newPlaylistInputData;
 import Database.playlistAccessInterface;
 import Database.playlistLibrary;
+
+//import OutputData.newPlaylistOutputData;
+//import OutputBoundary.newPlaylistOutputBoundary;
+
 public class createPlaylist implements newPlaylistInputBoundary {
+    //TODO: Do I need to create an output data instance attribute ? I dont know if this is a data structure
```
sarxhlwalker, line 13: You technically don't need an outputData attribute, as you're creating it and then immediately passing it to the OutputBoundary/Presenter.

src/main/java/UseCase/createPlaylist.java
```diff
@@ -5,26 +5,41 @@
...
+                     newplaylistInputData.getFirstSong());
         }
+        newplaylistInputData.getOwner().addPlaylist(playlist);
+        // save the newly created playlist to the RegularUser's playlist list
         library.savePlaylist(new playlistDsData(playlist));
+
+        // TODO: new potential implementation of sending confirmation message via outputdata, check correctness
+//        newPlaylistOutputData outputData = new newPlaylistOutputData("Playlist created!");
+//        this.newPlaylistOutputBoundary.playlistCreationConfirmation(outputData);
         return "Playlist created!";
```
sarxhlwalker, line 43: If I'm understanding your code + clean architecture correctly -- and if I'm not, let me know! -- I don't think you need to directly return the string. You should store the OutputBoundary/Presenter as an attribute and then call its function with the outputData (i.e. the string) as a parameter.

src/main/java/ViewModel/playlistViewModel.java
```diff
@@ -1,4 +1,37 @@
...
+import Presenter.playlistPresenter;
 public class playlistViewModel {
+    private final newPlaylistController newPlaylistController;
+    private  playlistPresenter playlistPresenter;
+
+    public playlistViewModel(int id, RegularUser user, String playlistName, Song firstSong){
+        // get necessary information from View to construct a new playlist
+        newPlaylistInputData inputData = new newPlaylistInputData(id,playlistName,firstSong,user);
+        //Blank output data object to extract success message from
+        newPlaylistOutputData outputData = new newPlaylistOutputData("");
```
sarxhlwalker, line 19: I don't think you should construct the OutputData in the presenter. I could be wrong, however.

sarxhlwalker: If that's the case, Presenter does not need to take OutputData as a parameter.

src/main/java/ViewModel/playlistViewModel.java
```diff
@@ -1,4 +1,37 @@
...
+        this.playlistPresenter = new playlistPresenter(this, outputData);
+//        newPlaylistInputBoundary newPlaylistInputBoundary = InputBoundary.newPlaylistInputBoundary;
+        createPlaylist newPlay= new createPlaylist();
```
sarxhlwalker, line 22: Similarly, I don't believe you should construct the use case interactor in the presenter. This should occur within the controller.

sarxhlwalker (commented):
I left a lot of comments, but please take what I wrote with a grain of salt. I may have misunderstood things based on our discussion today. I'm not sure whether the creation of InputData should occur in the ViewModel or the Controller (they're both in the same green layer), because my use case is so different. I'm mildly more confident in my other comments, however.

src/main/java/InputData/newPlaylistInputData.java
NicholasMacasaet: Will do

src/main/java/UseCase/createPlaylist.java
NicholasMacasaet: Thats true, I think I made my code significantly more complicated by trying to use output data

src/main/java/ViewModel/playlistViewModel.java
NicholasMacasaet: Ok, that's what I think was missing

src/main/java/UseCase/createPlaylist.java
sarxhlwalker: I understand that what i wrote might be a bit confusing! To try to make things clearer, i think what i'm envisioning is your createPlaylist stores the presenter/outputboundary as an instance attribute. then in the method, construct the outputdata, then call the presenter's function. you don't need to store the outputdata as an instance attribute itself, but you could if you want to.

src/main/java/Controller/newPlaylistController.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * @param presenter presenter object to store output message in
+     * @return String confirmation of the playlist's creation
+     */
+    public void createNewPlaylist(playlistPresenter presenter){
+        //TODO: Assuming this section functions like createPlaylist object w/ input data
+        // and setting the presenter's output message to the confirmation stored in aforementioned object
+        // which is accessed by showConfirmation()
+        this.newPlaylistInputBoundary.newPlaylist(this.newPlaylistInputData);
+        presenter.setOutputMessage(this.newPlaylistInputBoundary.showConfirmation());
```
sarxhlwalker, line 29: IIRC the diagram suggests the use case should take the presenter as a parameter (as an OutputBoundary) and call the presenter's function from within the use case interactor, but this technically does not break clean architecture. 

I'd prefer we keep the places where the presenter is called as consistent as possible, however.

src/main/java/InputData/newPlaylistInputData.java
```diff
@@ -1,22 +1,23 @@
...
-    public newPlaylistInputData(int id, String playlistName, User owner){
+    public newPlaylistInputData(int id, String playlistName, RegularUser owner){
         this.id = id;
         this.playlistName = playlistName;
         this.owner = owner;
         this.firstSong = null;
     }
 
-    public newPlaylistInputData(int id, String playlistName, Song firstSong, User owner){
+    public newPlaylistInputData(int id, String playlistName, Song firstSong, RegularUser owner){
```
sarxhlwalker, line 20: Similarly, you could add a comment here about why it's a RegularUser.

src/main/java/OutputData/editPlaylistOutputData.java
```diff
@@ -1,4 +1,22 @@
 package OutputData;
 
 public class editPlaylistOutputData {
+    String songAdded;
```
sarxhlwalker, line 4: is there a reason that these are default? if you have public getters, you could have these as private.

src/main/java/OutputData/newPlaylistOutputData.java
```diff
@@ -0,0 +1,11 @@
+package OutputData;
+public class newPlaylistOutputData {
+    String playlistCreated;
```
sarxhlwalker, line 3: same as above

src/main/java/UseCase/createPlaylist.java
```diff
@@ -5,27 +5,56 @@
 import InputData.newPlaylistInputData;
 import Database.playlistAccessInterface;
 import Database.playlistLibrary;
-public class createPlaylist implements newPlaylistInputBoundary {
+import OutputBoundary.newPlaylistOutputBoundary;
+import OutputData.newPlaylistOutputData;
+
+//import OutputBoundary.newPlaylistOutputBoundary;
```
sarxhlwalker, line 11: if you could remove commented out imports / functions, that would be great as it would make reading your code a lot easier.

src/main/java/UseCase/createPlaylist.java
```diff
@@ -5,27 +5,56 @@
...
+        //TODO #1: new potential implementation of sending confirmation message via outputdata, check correctness
+        //TODO #2: Regain will to live
+
+        //generate output data
+        newPlaylistOutputData outputData = new newPlaylistOutputData("Playlist created!");
+        // this.newPlaylistOutputBoundary.playlistCreationConfirmation(outputData);
+        // this.outputBoundary.getPlaylistCreationConfirmation(outputData);
+
+        //set message attribute to output data
+        this.message = this.outputBoundary.getPlaylistCreationConfirmation(outputData);
```
sarxhlwalker, line 53: why is the outputBoundary / presenter returning information to the use case? it should be fine to call the presenter's function and not store the resulting message in the use case interactor.

sarxhlwalker (changes requested):
Looks good! Please see the outstanding comments.

src/main/java/OutputData/editPlaylistOutputData.java
NicholasMacasaet: The edit playlist stuff I created just to have something implemented, I updated those variable to be private now.

src/main/java/Controller/newPlaylistController.java
NicholasMacasaet: That method I changed in the most recent commit I'm going to push now, so it should be better.

NicholasMacasaet:
Going to go to Paul Gries' office hours tomorrow to get feedback on my use case, if everything is good I'll implement editPlaylist almost exactly how I implemented this use case.

sarxhlwalker:
> Going to go to Paul Gries' office hours tomorrow to get feedback on my use case, if everything is good I'll implement editPlaylist almost exactly how I implemented this use case.

Thanks for the quick updates. I'll re-review this ASAP, but for ease of reviewing, could you either implement editPlaylist in a different branch / PR, or wait to add those commits to this branch until this has been merged? I expect / hope we'll be able to merge this into main soon, because it's been open for a while now.

NicholasMacasaet:
> > Going to go to Paul Gries' office hours tomorrow to get feedback on my use case, if everything is good I'll implement editPlaylist almost exactly how I implemented this use case.
> 
> Thanks for the quick updates. I'll re-review this ASAP, but for ease of reviewing, could you either implement editPlaylist in a different branch / PR, or wait to add those commits to this branch until this PR has been merged? I expect / hope we'll be able to merge this into main soon, because it's been open for a while now.

If you do do some more reviewing could you wait until tomorrow if possible ? (check your discord for said reason). And of course I'll wait until this branch is merged. In regards to the length of time this has been open that is my fault due to my previously lacking understanding with how the view model and presenter stuff fit in.

NicholasMacasaet:
After getting feedback from Gries he said my code was pretty good and all I needed to do was just move some stuff around. I'm ready to code edit playlist when this goes through

src/main/java/Controller/newPlaylistController.java
```diff
@@ -0,0 +1,48 @@
...
+ * Controller-layer implementation for create playlist use case
+ */
+public class newPlaylistController {
+    private final newPlaylistInputBoundary newPlaylistInputBoundary;
+
+//    private final newPlaylistInputData newPlaylistInputData;
+//
+//    private CreatePlaylistInteractor createPlaylist;
+
+    public newPlaylistController(CreatePlaylistInteractor createPlaylist){
```
chrstinalin, line 19: This isn't technically wrong, but for the sake of consistency I believe we were making the controller construct the interactor.

chrstinalin: You'd probably also want to take in the Presenter as an OutputBoundary (as you had commented out originally) to give to the Interactor

src/main/java/Presenter/playlistPresenter.java
```diff
@@ -1,4 +1,42 @@
...
+import OutputBoundary.newPlaylistOutputBoundary;
+import OutputData.newPlaylistOutputData;
+import ViewModel.playlistViewModel;
 
-public class playlistPresenter {
+public class playlistPresenter implements newPlaylistOutputBoundary {
+
+    private String outputMessage;
+
+    final playlistViewModel viewModel;
```
chrstinalin, line 10: RE: Sarah

src/main/java/UseCase/CreatePlaylistInteractor.java
```diff
@@ -0,0 +1,49 @@
...
+import Database.playlistAccessInterface;
+import Database.playlistLibrary;
+import OutputBoundary.newPlaylistOutputBoundary;
+import OutputData.newPlaylistOutputData;
+
+public class CreatePlaylistInteractor implements newPlaylistInputBoundary {
+        private newPlaylistOutputBoundary outputBoundary;
+//        private String message;
+    public CreatePlaylistInteractor(newPlaylistOutputBoundary outputBoundary){
+        this.outputBoundary = outputBoundary;
```
chrstinalin, line 15: Not wrong at all but if you wanted to make this clearer to yourself you could rename outputBoundary to Presenter (since thats what would be passed).

src/main/java/UseCase/CreatePlaylistInteractor.java
```diff
@@ -0,0 +1,49 @@
...
+import Entities.Playlist;
+import InputBoundary.newPlaylistInputBoundary;
+import InputData.newPlaylistInputData;
+import Database.playlistAccessInterface;
+import Database.playlistLibrary;
+import OutputBoundary.newPlaylistOutputBoundary;
+import OutputData.newPlaylistOutputData;
+
+public class CreatePlaylistInteractor implements newPlaylistInputBoundary {
+        private newPlaylistOutputBoundary outputBoundary;
```
chrstinalin, line 12: private final?

src/main/java/UseCase/CreatePlaylistInteractor.java
```diff
@@ -0,0 +1,49 @@
...
+//        this.message = "";
+    }
+    /**
+     * @param newplaylistInputData holds necessary data to instantiate a new playlist
+     * if hasFirstSong() is false, then newPlaylist() will instantiate an empty playlist and vice versa if it's value
+     * is true
+     */
+    public void newPlaylist(newPlaylistInputData newplaylistInputData) {
+        Playlist playlist;
+        playlistAccessInterface library = playlistLibrary.getInstance();
```
chrstinalin, line 25: (Mostly for consistency) Add a `private final playlistAccessInterface library;` up top and move line 25 to the initializer

src/main/java/UseCase/editPlaylist.java
```diff
@@ -10,7 +10,9 @@ public class editPlaylist implements editPlaylistInputBoundary{
      *                  and changeName()
      *
     */
```
chrstinalin, line 12: Initializer?

src/main/java/UseCase/editPlaylist.java
```diff
@@ -20,6 +22,7 @@ public String removeSong(editPlaylistInputData inputData) {
         }
         return inputData.getSong().getName()+" removed ";
     }
+
     // TODO: Work on view later
     public String addSong(editPlaylistInputData inputData){
```
chrstinalin, line 27: Instead of returning a String directly, you'd have passed in a Presenter in the initializer whose method you'd call

chrstinalin (changes requested):
See comments

src/main/java/UseCase/editPlaylist.java
NicholasMacasaet: Ignore this for now, I was going to correctly code it later

src/main/java/UseCase/editPlaylist.java
NicholasMacasaet: This is true, I havent coded that part as I was more focused on correctly coding the create playlist use case, then updating edit playlist to do the same

src/main/java/Controller/newPlaylistController.java
NicholasMacasaet: Ah, I might have missed that then when re-working my code

NicholasMacasaet: That would require creating a new playlist interactor object in the constructor and setting this.inputboundary to said object correct ?

NicholasMacasaet: Also Gries' code does something similar but instead passes an input boundary object in the constructor instead of a createplaylist interactor

src/main/java/InputBoundary/newPlaylistInputBoundary.java
```diff
@@ -2,5 +2,9 @@
 import InputData.newPlaylistInputData;
 
 public interface newPlaylistInputBoundary {
-    public String newPlaylist(newPlaylistInputData inputData);
+    void newPlaylist(newPlaylistInputData inputData);
+
+    default String showConfirmation(){
```
rafaelgacesa, line 7: This should not be a default method - this means that things that implement this interface can use this 'default' version of the method, which would be unwanted.

Instead, you should have it be public, which will also require replacing the braces with a semicolon.

src/main/java/Controller/newPlaylistController.java
```diff
@@ -0,0 +1,46 @@
...
+import InputData.newPlaylistInputData;
+/**
+ * Controller-layer implementation for create playlist use case
+ */
+public class newPlaylistController {
+    private final newPlaylistInputBoundary newPlaylistInputBoundary;
+
+    /**
+     *
+     * @param createPlaylistInteractor interactor object to be passed into controller for use case
```
rafaelgacesa, line 14: This is assuming that we will be passing an interactor into the controller, which is not the way we've been working in previous controllers (thus far we have been constructing them inside the controller). The other option is to construct the interactor in the view model, which it appears is what you are doing. While this is valid, and will work (and might even be the better choice tbh, now that I think about it), perhaps we should talk about this as a group and decide on one method or the other for consistency across our codebase.

rafaelgacesa: For now, no changes need to be made - it works and is a 'correct' implementation by clean architecture standards. I think I'll even lobby for changing our other controller implementations to follow this one - it's less clutter since we don't have to ever give the controller the presenter. Thanks for making me realize this lol

src/main/java/UseCase/CreatePlaylistInteractor.java
```diff
@@ -0,0 +1,51 @@
...
+        Playlist playlist;
+        if (!newplaylistInputData.hasFirstSong()) {
+            playlist = new Playlist(newplaylistInputData.getId(), newplaylistInputData.getPlaylistName(),
+                    newplaylistInputData.getOwner());
+        }
+        else{
+            playlist = new Playlist(newplaylistInputData.getId(), newplaylistInputData.getPlaylistName(), newplaylistInputData.getOwner(),
+                     newplaylistInputData.getFirstSong());
+        }
+        newplaylistInputData.getOwner().addPlaylist(playlist);
```
rafaelgacesa, line 38: This is a great implementation and correct use of the playlist database. However, I am wondering if you will also have to use the user database class to save the user, after assigning them the new playlist.

I think you will have to do this, since there is no connection between the RegularUser class and the user database to say, "Hey, I have a new playlist that needs to be added to my data and saved!". So, I think you'll have to add the instance variable `userAccessInterface userDatabase` to the top of your file, and then instantiate it similarly to the playlist library, and then call the `userDatabase.save()` method on the `RegularUser` after you add the playlist to it in line 38. This will ensure that the new playlist is correctly written to the database as belonging to the user.

src/main/java/ViewModel/playlistViewModel.java
```diff
@@ -1,4 +1,44 @@
...
-
+import Presenter.playlistPresenter;
 public class playlistViewModel {
+    private playlistPresenter playlistPresenter;
+    private String outputMessage;
+
+    public playlistViewModel(){
+        // get necessary information from View to construct a new playlist
+        //TODO: This is the controller, just pass this to the controller, move to controller
+        //TODO: Create controller once in the view as an instance variable / pass as a parameter in main
```
rafaelgacesa, line 10: This is the method Gries likes, since he really focuses on dependency injection. However, we've been created the controllers and presenters inside of the view model, which is definitely allowed according to clean architecture. I've also personally asked Prof. Gries about this and he said that it is fine to do this too, he just prefers dependency injection. 

I think we should stick to constructing our controllers and presenters inside of the view model, just to keep our code consistent, even though it introduces some hard dependencies. If we have time at the end of development, we can take a step back and see if we can apply dependency injection anywhere to make things cleaner, but for now we should be consistent.

rafaelgacesa (changes requested):
Overall looks like good progress on your use case, no major errors. Some minor comments regarding small changes and things to think about.

rafaelgacesa (approved):
Thanks for implementing the changes! I got a bit confused about the input boundary (since it's empty now) but I realized that's the one for the use case you haven't implemented yet. This pull looks good to go!

Mistryr222 (approved):
Looks Good!

chrstinalin (approved):
Looks good!

PR #41, opened 2022-11-12
Feature 7 - playSong() - needs to be tested, minor changes and additions to some of the skeletal files

Mistryr222:
1. playSong() implementation, in the use case there is a small issue with playAudio being called from playSong's use case, but its like a one line fix
2. Changes and additions to many of the skeletal files we had before
3. shufflePlaylist() implementation

### DIFF
--- a/src/main/java/Entities/Playlist.java
+++ b/src/main/java/Entities/Playlist.java
@@ -1,10 +1,10 @@
 package Entities;
 
-import java.util.ArrayList;
 import java.util.LinkedList;
 
 /**
  * Entities layer class representing a Playlist.
+ * TODO Edge Case: song is removed while its being played should not affect the songs being played, or playlists
  */
 public class Playlist {
     private String name;
@@ -40,7 +40,6 @@ public class Playlist {
     public String getName(){
         return this.name;
     }
-    // no set owner because the owner is permanently set upon initialization
 
     /**
      * @return User object representing owner of a playlist
--- a/src/main/java/Entities/Song.java
+++ b/src/main/java/Entities/Song.java
@@ -8,7 +8,7 @@ public class Song {
 
     private final int id;
     private final String name;
-    private final List artistList;
+    private final List<String> artistList;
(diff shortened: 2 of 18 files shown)


Mistryr222:
@sarxhlwalker your use of playAudio() needs to change now since @rafaelgacesa's MusicPlayer class takes in a Song instance rather than a file

rafaelgacesa:
Something's broken in your playAudio, something about assigning a final variable? I can't tell exactly because I'm on mobile.

src/main/java/Entities/Song.java
```diff
@@ -8,7 +8,7 @@ public class Song {
 
     private final int id;
     private final String name;
-    private final List artistList;
+    private final List<String> artistList;
```
Mistryr222, line 11: Also I didn't know whether these should be ArtistUsers or Strings so let me know if changes need to be made on Monday

Mistryr222:
> Something's broken in your playAudio, something about assigning a final variable? I can't tell exactly because I'm on mobile.

Yes it's because of something related to some static stuff, we can fix it on Monday's meeting once we test out playAudio().

src/main/java/Entities/Playlist.java
```diff
@@ -1,10 +1,10 @@
 package Entities;
 
-import java.util.ArrayList;
 import java.util.LinkedList;
 
 /**
  * Entities layer class representing a Playlist.
+ * TODO Edge Case: song is removed while its being played should not affect the songs being played, or playlists
```
NicholasMacasaet, line 7: Nice catch, will need to work out the kinks with this later when the front-end stuff is more developed.

src/main/java/Entities/Song.java
NicholasMacasaet: I would assume its a list of ArtistUsers based off of the context, then you could just modify the getArtistList() method to go through the  artistList and retrieve the String representation of the artistUser's names and return a list of those

PR #42, opened 2022-11-12
Feature 2- register

SiqiZhang6:
Made the userController and merged the inputboundary for login and register (since they share the same data and boundary). The red in userController.java is because loginIB have not been implemented.


gradlew
```diff
@@ -0,0 +1,234 @@
+#!/bin/sh
```
sarxhlwalker, line 1: Could you remove gradle/wrapper, gradlew, gradlew.bat, and .idea files from the PR?

sarxhlwalker (changes requested):
The controller looks good, and removing registerInputBoundary makes sense. I don't think we should have the .idea and gradle files in our repo. Is the reason the auto-grade fails because the input boundary isn't implemented?

gradlew
SiqiZhang6: Yep I tried deleting them, but the gradle/wrapper/gradle-wrapper.jar did not have the delete option :(. And yes I assumed that the tests failed because loginIB was not implemented.

src/main/java/Controller/userController.java
```diff
@@ -1,4 +1,37 @@
...
 
+import InputBoundary.loginInputBoundary;
+import InputData.loginInputData;
+
+
+/**
+ * Controller for login and register use cases.
+ */
 public class userController {
+    final loginInputBoundary loginInteractor, registerInteractor;
```
manya-mittal, line 11: Shouldn't these be private? I think if you don't specify then it will be made public by the compiler but it should be private. Could you please add the private keyword? Can you also please make these on seperate lines please?

src/main/java/Controller/userController.java
```diff
@@ -1,4 +1,37 @@
...
+    public userController(loginInputBoundary loginInteractor, loginInputBoundary registerInteractor) {
+        this.loginInteractor = loginInteractor;
+        this.registerInteractor = registerInteractor;
+    }
+
+    /**
+     * function calling the use case for playing the space
+     * @param loginID input data for playing the space
+     */
+    public void login(loginInputData loginID){
```
manya-mittal, line 26: This inputData should be made in the controller rather than being passed in. So the controller class should have an inputData instance variable that is initialised in the constructor. You can then pass in the instance variable into your method

manya-mittal (changes requested):

PR #43, opened 2022-11-13
Song Database Implementation: songLibrary, songDsData, bits and pieces.

chrstinalin:
Title -- Pull includes the implementation of everything relating to Song's database: mainly songLibrary and songDsData. Brief summary;

**CHANGES**

- **Uploaded the standard library of songs.** Currently located in the new folder songLib under src. Any test cases or other pieces of code that currently create new instances of Songs should likely be re-reviewed with this in mind. I already fixed a few, but I might've missed some.
- Updated build.gradle to include jaudiotagger
- Updated Song entity to include username of uploading user.
- Reading/saving from songs.csv, formatted as `ID, uploader, filepath`
- Changed artistList from List<String> to String[]
- Removed 'length' as Jaudiotagger cannot retrieve it. If we want to show it, it would be on the Jlayer side. Better suited this way, anyway.
- Removed isExplicit. Too much of a pain for too little gain.
- changed saveSong to return a boolean for a successful song addition.
- Created Test file for SongLibrary.

**KNOWN ISSUES**

- createFile() assumes the existence of a user admin, as it assigns all songs currently in songLib /to/ admin. (This was only important for creating songs.csv from scratch. It won't do this now that it exists.)

- Many files don't have album covers. I'll be creating default covers to put in the BufferedImage parameter later. I don't think anyone is at that stage (which is why I'm putting it off for a later PR), but please don't try accessing the BufferedImage parameter until I do.

- ~~When parsing ID names, the 0s at the beginning of the names are omitted. Theoretically, this shouldn't create duplicate IDs anyway, as 1) what's being checked is the parsed ID, and 2) randInt will not create more IDs that start with 0. I'll go back and rename the files to omit the 0s (not that difficult), but I figure its lower priority due to the reasons I stated.~~ Fixed 2022/11/14

- Need to implement safeguard against improperly formatted mp3s (ex. missing genre). There currently aren't any, so it should be temporarily OK, but I have some code smells because of this.

Making this a PR despite this so you guys have a better SONG_LIBRARY to work with for now.


chrstinalin:
Looks like there's issues with autograding accessing files (I think Raf might have mentioned this before...?) since these tests are working fine for me.

rafaelgacesa:
> Looks like there's issues with autograding accessing files (I think Raf might have mentioned this before...?) since these tests are working fine for me.
> 
> Not sure about the Playlist ones, though.

Yep, I’ve been commenting out assertions for any unit tests requiring file I/O, so when they merge it doesn’t break autograder for every future pull.

chrstinalin:
> > Looks like there's issues with autograding accessing files (I think Raf might have mentioned this before...?) since these tests are working fine for me.
> > Not sure about the Playlist ones, though.
> 
> Yep, I’ve been commenting out assertions for any unit tests requiring file I/O, so when they merge it doesn’t break autograder for every future pull.

Hm, OK -- I guess I'll do that too.

rafaelgacesa:
Looks like it’s my tests failing - I’d bet this is because I was using random song IDs that don’t actually exist with that test implementation of songLibrary. Do you want to take a look at this or should I at a later time?

chrstinalin:
> Looks like it’s my tests failing - I’d bet this is because I was using random song IDs that don’t actually exist with that test implementation of songLibrary. Do you want to take a look at this or should I at a later time?

Hm, but if I/O isn't working at all this shouldn't really be a problem, no?

rafaelgacesa:
> > Looks like it’s my tests failing - I’d bet this is because I was using random song IDs that don’t actually exist with that test implementation of songLibrary. Do you want to take a look at this or should I at a later time?
> 
> Hm, but if I/O isn't working at all this shouldn't really be a problem, no?

I think the issue would be that somewhere in playlistLibrary (and maybe userList for ArtistUser) I’m requesting IDs that don’t exist, which would throw a null pointer ? Check playlists.csv and artists.csv, they’re going to have a bunch of random IDs.

chrstinalin:
> > > Looks like it’s my tests failing - I’d bet this is because I was using random song IDs that don’t actually exist with that test implementation of songLibrary. Do you want to take a look at this or should I at a later time?
> > 
> > Hm, but if I/O isn't working at all this shouldn't really be a problem, no?
> 
> I think the issue would be that somewhere in playlistLibrary (and maybe userList for ArtistUser) I’m requesting IDs that don’t exist, which would throw a null pointer ? Check playlists.csv and artists.csv, they’re going to have a bunch of random IDs. 


That would make sense for the NullPointerException, but the majority seem to be NoClassDefFound which is a bit odd.

NicholasMacasaet:
>

> I took a look at the error it's throwing, check to see if the class path is set to the location where the class loader is,

chrstinalin:
> I took a look at the error it's throwing, check to see if the class path is set to the location where the class loader is. Did you check if main is set as sources root ?

Would it being marked as such affect the Github result? I can try, but I don't think I'd be able to push it just like that.

rafaelgacesa (approved):
**Also in this pull:**
- fixed some issues with `userList` incorrectly modifying `artists.csv`
- rebuilt the autograder to be more robust (unnecessarily lol)
- added back playAudio use case that was somehow removed?
- added back `userList` testing

I have personally reviewed almost every file here, everything else looks and works fine.

Mistryr222 (approved):
Database features look good. Nice organization of the work.

sarxhlwalker (approved):
Great code! Thanks for explaining the reason behind removing the length attribute.

PR #49, opened 2022-11-15
PlayPlaylist: controller, presenter, view model

rafaelgacesa:
### In this pull:
- the beginnings of the `musicEngineController`, completed and functional for the play playlist use case
- the beginnings of the `songPresenter`, completed and functional for the play playlist use case
- basic view model that will be responsible for any use case playing audio
- `IDButton`, a button which contains a field to store the ID of a linked song or playlist
- modifications to `playlistInputData` to include a field for ID


Mistryr222 (dismissed):
playPlaylist() use case looks good, we will need to override them in my pull.

Mistryr222 (approved):
Music Engine Controller seems to use the use cases correctly, well done.

sarxhlwalker (approved):
Looks good! Thanks for add the IDButton.

PR #50, opened 2022-11-15
Feature 2 register user

SiqiZhang6:
Register user usecase updates:
- realized that input boundaries should be separate, so I made the IB for both login and register
- separated the usercontroller into logincontroller and registercontroller, as well as added the RegisterArtistController
- separated the register usecase into two usescases (so we can still have artistname), each responsible for registering regular/artist users.*
- implemented the inputdata and inputboundaries for login, register and register artist

New stuff that comes with new usecase:
- RegisterArtistInteractor,  usecase creating an artist user
- RegisterArtistInputBoundary, RegisterArtistInputData

The next step is to implement the output part and the presenter.

p.s apologies for all the file changes


src/main/java/UseCase/RegisterInteractor.java
```diff
@@ -0,0 +1,46 @@
...
+    public RegisterInteractor(loginOutputBoundary registerPresenter, userAccessInterface Allusers, userFactory userFactory){
+        this.Allusers = Allusers;
+        this.registerPresenter =registerPresenter;
+        this.userFactory = userFactory;
+    }
+    @Override
+    public void register(loginInputData loginID){
+        if (Allusers.exists(loginID.getUsername())){
+            //TODO: prepare user exists failure view
+        } else if (loginID.getPassword() == "") {
```
chrstinalin, line 26: The username field could also be empty.

SiqiZhang6: Ah yes I will add that in, thanks

src/main/java/UseCase/RegisterInteractor.java
```diff
@@ -0,0 +1,46 @@
...
+            //TODO: prepare no password failure view
+        }
+
+        if (loginID.isArtist()){
+            User user = this.userFactory.createArtistUser("", loginID.getUsername(), loginID.getPassword());
```
chrstinalin, line 31: Hold up -- why is artistName being initialized as ""? The variable within the ArtistUser entity is final anyway, so it's not capable of being changed at a later point.

You might need to make a separate ArtistInputData (or some other name) for registering artists, so we can get that name.

src/main/java/UseCase/RegisterInteractor.java
```diff
@@ -0,0 +1,46 @@
...
+            userDsData newUser = new userDsData(user);
+            Allusers.save(newUser);
```
chrstinalin, line 33: Redundant but since I'm here, you could simplify lines 32 and 33 into simply ` Allusers.save(new userDsData(user));`. Not that important though, mostly personal preference.

chrstinalin (changes requested):
See comments -- I'm mainly concerned with the issue relating to artistName.

src/main/java/UseCase/RegisterInteractor.java
SiqiZhang6: the thing is that I thought we dont take in artist name at login? so do u mean like adding an extra input for artist registration?

src/main/java/UseCase/RegisterInteractor.java
SiqiZhang6: noted, thx. I'll change accordingly :)

src/main/java/Controller/userController.java
```diff
@@ -1,4 +1,42 @@
...
+    private final loginInputData loginInputData;
+    private final loginInputData registerInputData;
+    /**
+     * Constructor.
+     * @param loginInteractor use case interactor for adding a song to the space
+     * @param registerInteractor use case interactor for playing the space
+     */
+    public userController(loginInputBoundary loginInteractor, RegisterInputBoundary registerInteractor,
+                          String username, String password, boolean isArtist) {
+        this.loginInteractor = loginInteractor;
```
sarxhlwalker, line 23: IIRC the controller should construct the loginInteractor -- your code implies that the ViewModel creates it. Technically, it's fine if the ViewModel does this (as they're in the same layer), but I'd prefer we stay consistent across use cases. 

Same applies for registerInteractor.

Thanks for adding Interactor as a suffix to the use case files, btw!

sarxhlwalker (commented):
Mostly looks good to me, except one comment. 

Btw @chrstinalin could you confirm that it's OK that gradle-wrapper.jar is being created + pushed to main in this PR? IIRC your PR did as well.

src/main/java/UseCase/RegisterInteractor.java
SiqiZhang6: > If the password field can be left empty, I presume the username field could also be left empty.
> 
> Just an aside, are we having minimum characters for each? I guess it would add an unnecessary layer of complexity at the "prototype" stage.

lol i did not read the whole comment. the fail view is presented if username/password is empty so no they can't be empty. Also as of now we don't have the minimum characters, we can add that in later if u want?

src/main/java/Controller/userController.java
SiqiZhang6: Ic, I'll add that in rn. Thanks!

chrstinalin:
> Btw @chrstinalin could you confirm that it's OK that gradle-wrapper.jar is being created + pushed to main in this PR? IIRC your PR did as well.

Do you mean my closed one? I don't remember selecting it (it might've been a result of Raf trying to fix the tests in my branch?), but I think it should be fine.

src/main/java/Controller/LoginController.java
```diff
@@ -0,0 +1,27 @@
...
+ * Controller for login and register use cases.
+ */
+public class LoginController {
+    private final loginInputBoundary loginInteractor;
+    private final loginInputData loginInputData;
+    /**
+     * Constructor.
+     * @param loginInteractor use case interactor for user logging in
+     */
+    public LoginController(loginInputBoundary loginInteractor, String username, String password, boolean isArtist) {
```
chrstinalin, line 17: The initializer should be passed a presenter, which is then passed to the interactor.

The controller initializer should also construct the loginInteractor (Just for the sake of consistency, as Sarah previously mentioned) rather than be passed it.

The ViewModel would be the one creating the InputData object, passed directly to the login() rather than to the initializer -- I imagine creating new instances of LoginController for each login attempt would be problematic.

src/main/java/Controller/LoginController.java
```diff
@@ -0,0 +1,27 @@
...
+        this.loginInteractor = loginInteractor;
+        this.loginInputData = new loginInputData(username, password, isArtist);
+    }
+    /**
+     * function calling the use case for logging in
+     */
+    public void login(){
+        this.loginInteractor.login(this.loginInputData);
```
chrstinalin, line 25: See previous comment -- login should be taking in InputData and passing it on.

src/main/java/Controller/LoginController.java
```diff
@@ -0,0 +1,27 @@
...
+import InputBoundary.loginInputBoundary;
+import InputData.loginInputData;
+
+
+/**
+ * Controller for login and register use cases.
+ */
+public class LoginController {
+    private final loginInputBoundary loginInteractor;
+    private final loginInputData loginInputData;
```
chrstinalin, line 12: If you recall from lecture (I don't blame you if you don't it was ages ago), Input/OutputData objects are meant to be temporary.

src/main/java/Controller/RegisterController.java
```diff
@@ -0,0 +1,41 @@
...
+import InputData.loginInputData;
+
+
+/**
+ * Controller for login and register use cases.
+ */
+public class RegisterController {
+    private final RegisterInputBoundary registerInteractor;
+    private final RegisterArtistInputBoundary artistInteractor;
+    private final RegisterInputData registerInputData;
```
chrstinalin, line 17: Same point as in loginController -- registerInputData should be temporary.

Plus, since the Controller is final in the ViewModel, this would also never change -- meaning that, if a user were to incorrectly register or log in, their incorrect attempt would be the final registerInputData and incapable of being replaced.

src/main/java/Controller/RegisterController.java
```diff
@@ -0,0 +1,41 @@
...
+    private final RegisterArtistInputData artistInputData;
+    /**
+     * Constructor.
+     * @param registerInteractor use case interactor for creating regular users
+     */
+    public RegisterController(RegisterInputBoundary registerInteractor, RegisterArtistInputBoundary artistInteractor,
```
chrstinalin, line 23: See initializer comment in loginController

src/main/java/InputData/RegisterArtistInputData.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * @return the password of user
+     */
+    public String getPassword() {
+        return this.password;
+    }
+    /**
+     * @return whether the user is an artist
+     */
+    public boolean isArtist() {
```
chrstinalin, line 36: Since we're splitting RegisterArtist and Register, is this still necessary?

We're keeping ArtistUser and RegularUser as separate entities now, correct? Then we can just use an 'is a' check (if it were to come up) rather than an isArtist parameter.

src/main/java/InputData/loginInputData.java
```diff
@@ -1,4 +1,31 @@
...
+    /**
+     * @return the password of user
+     */
+    public String getPassword() {
+        return this.password;
+    }
+    /**
+     * @return whether the user is an artist
+     */
+    public boolean isArtist() {
```
chrstinalin, line 28: RE: my comment in RegisterArtistInputData, I figure this would still be needed in _this_ case (since the two types of users are sharing on this end, it'd be important to distinguish between the two later on).

In RegisterArtistInputData and RegisterInputData, though, they each have their own separate use cases -- so whichever one is being called would implicitly tell us the type.

src/main/java/UseCase/RegisterArtistInteractor.java
```diff
@@ -0,0 +1,41 @@
...
+import Entities.User;
+import Entities.userFactory;
+import InputBoundary.RegisterArtistInputBoundary;
+import InputData.RegisterArtistInputData;
+import OutputBoundary.loginOutputBoundary;
+
+public class RegisterArtistInteractor implements RegisterArtistInputBoundary{
+    private final loginOutputBoundary registerPresenter;
+    private final userFactory userFactory;
+
```
chrstinalin, line 14: You haven't implemented this yet which is probably why you don't have it, but in case you want to know how to access the database -- 
- Add `private final userAccessInterface userList;` here
- In the initializer, `this.userList = Database.UserList.getInstance();`

src/main/java/UseCase/RegisterArtistInteractor.java
```diff
@@ -0,0 +1,41 @@
...
+    userAccessInterface allUsers;
```
chrstinalin, line 15: In case you weren't sure how to access the database (considering it isn't initialized --)
- Change to `private final userAccessInterface userList;` (or keep it as allUsers, doesn't really matter)
- In the initializer, `this.userList = Database.UserList.getInstance();`

src/main/java/UseCase/RegisterArtistInteractor.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public RegisterArtistInteractor(loginOutputBoundary registerPresenter, userAccessInterface allUsers,
+                                    userFactory userFactory){
+        this.allUsers = allUsers;
```
chrstinalin, line 19: See above comment.

src/main/java/UseCase/RegisterArtistInteractor.java
```diff
@@ -0,0 +1,41 @@
...
+        this.registerPresenter =registerPresenter;
+        this.userFactory = userFactory;
```
chrstinalin, line 21: I'm not certain if userFactory would be passed in like this, actually. @sarxhlwalker I know you removed spaceFactory, but are you more familiar with this?

src/main/java/UseCase/RegisterArtistInteractor.java
```diff
@@ -0,0 +1,41 @@
...
+            //TODO: prepare user exists failure view
+        } else if (registerInputData.getPassword() == "") {
+            //TODO: prepare no password failure view
+        } else if (registerInputData.getUsername() == "") {
+            //TODO: prepare no username failure view
+        }
+
+        //create artist user
+        User user;
+        user = this.userFactory.createArtistUser(registerInputData.getArtistName(),registerInputData.getUsername(),
```
chrstinalin, line 35: Lines 34 and 35 can be merged.

src/main/java/UseCase/RegisterInteractor.java
```diff
@@ -0,0 +1,40 @@
...
+import InputBoundary.RegisterInputBoundary;
+import Database.userAccessInterface;
+import Database.userDsData;
+import Entities.userFactory;
+
+public class RegisterInteractor implements RegisterInputBoundary{
+    private final loginOutputBoundary registerPresenter;
+    private final userFactory userFactory;
+
+    userAccessInterface allUsers;
```
chrstinalin, line 15: Same database comment as in RegisterArtistInteractor.

chrstinalin (changes requested):
See comments. Also, I found it helpful to work on the files in order of how the program would have traversed them itself (ViewModel, Controller, Interactor, Presenter, back to ViewModel) to make sense of _why_ things are set up the way they are.

src/main/java/UseCase/RegisterArtistInteractor.java
sarxhlwalker: I'm not super familiar, but I think it would make more sense to construct a new userFactory within the interactor (as technically userFactory is an entity, so I don't believe you should be creating a new instance of an entity outside of the red layer). it doesn't take any construction parameters, so it wouldn't really change anything, I don't think.

src/main/java/InputData/loginInputData.java
SiqiZhang6: Yes that's why I kept it, I had the exact same train of thought

src/main/java/UseCase/RegisterArtistInteractor.java
SiqiZhang6: They are now updated, thanks!

src/main/java/UseCase/RegisterArtistInteractor.java
SiqiZhang6: they r now merged. They were separate before because interactor was shared.

src/main/java/UseCase/RegisterArtistInteractor.java
```diff
@@ -0,0 +1,40 @@
...
+import InputData.RegisterArtistInputData;
+import OutputBoundary.loginOutputBoundary;
+
+public class RegisterArtistInteractor implements RegisterArtistInputBoundary{
+    private final loginOutputBoundary registerPresenter;
+    private final userFactory userFactory;
+
+    private final userAccessInterface allUsers;
+
+    public RegisterArtistInteractor(loginOutputBoundary registerPresenter, userAccessInterface allUsers,
```
NicholasMacasaet, line 17: Why are you passing an interface in the interactor's constructor?

src/main/java/UseCase/RegisterInteractor.java
```diff
@@ -0,0 +1,41 @@
...
+import Database.userDsData;
+import Entities.userFactory;
+
+public class RegisterInteractor implements RegisterInputBoundary{
+    private final loginOutputBoundary registerPresenter;
+    private final userFactory userFactory;
+
+    private final userAccessInterface allUsers;
+
+    public RegisterInteractor(loginOutputBoundary registerPresenter, userAccessInterface allUsers,
```
NicholasMacasaet, line 18: Similar to my previous comment why are you passing an interface into the constructor?

NicholasMacasaet (commented):
See comments

NicholasMacasaet (changes requested):
Probably remove the userAccessInterface in both of your interactors since you can just initilaize the userAccessInterface allUsers attribute inside of the constructors.

NicholasMacasaet (dismissed):
Everything looks good.

src/main/java/Controller/LoginController.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Controller for login and register use cases.
+ */
+public class LoginController {
+    loginInputBoundary loginInteractor;
+    /**
+     * Constructor.
+     * @param loginInteractor use case interactor for user logging in
+     */
+    public LoginController(loginInputBoundary loginInteractor) {
```
sarxhlwalker, line 16: As we discussed just earlier today, please don't pass the interactor into the controller; the controller should construct the interactor and the input data.

src/main/java/Controller/RegisterArtistController.java
```diff
@@ -0,0 +1,24 @@
...
+import OutputBoundary.loginOutputBoundary;
+
+/**
+ * Controller for login and register use cases.
+ */
+public class RegisterArtistController {
+    RegisterArtistInputBoundary registerInteractor;
+
+    public RegisterArtistController(RegisterArtistInputBoundary registerInteractor) {
+        this.registerInteractor = registerInteractor;
```
sarxhlwalker, line 15: same as above

src/main/java/Controller/RegisterController.java
```diff
@@ -0,0 +1,23 @@
...
+
+
+/**
+ * Controller for register use cases.
+ */
+public class RegisterController {
+    RegisterInputBoundary registerInteractor;
+
+    public RegisterController(RegisterInputBoundary registerInteractor) {
+        this.registerInteractor = registerInteractor;
```
sarxhlwalker, line 14: same as above

sarxhlwalker (changes requested):
minor changes

src/main/java/UseCase/RegisterArtistInteractor.java
SiqiZhang6: userFactory is now constructed in the constructer, thanks for the comment

sarxhlwalker (approved):
perfect thank u amazing

NicholasMacasaet (approved):

chrstinalin (approved):
Very nice, thanks for making those changes!!

src/main/java/UseCase/RegisterInteractor.java
chrstinalin: What I had meant was that, initially, having no username entered but a password entered would have passed a success view. Probably should have been more concise.

PR #57, opened 2022-11-16
uploadSong Use Case, songLibrary fixes, split songAccessInterface into two separate Data Access Interfaces

chrstinalin:
**CHANGELOG**

- Split songAccessInterface into two AccessInterfaces (as per Gries' lecture today -- (likely butchered tl;dr: AccessInterfaces should have one purpose, but the actual Database access class can implement multiple of them at once.) Refactored accordingly.

- Moved the backslash issue fix Raf made in regards to SONG_LIBRARY's filepath directly into the initializer.

- Implemented a non-specific safeguard mentioned in the 'known issues' of my https://github.com/CSC207-2022F-UofT/course-project-team-communify/pull/43#issue-1447018800.

- Updated saveSong() to actually be useful in regards to my use case.

- Implemented UploadSongInteractor

- Implemented a rough approximation of UploadSongPresenter, uploadSongController, and uploadSongViewModel to make sense of how clean architecture is traversed. Very rough and likely missing extra bits and pieces -- any corrections here would be greatly appreciated :)

- Added uploadSongTest skeleton.

- Added getString() to songDsData to return a String[] representation of the contained Song.

- Added getString() to the songLibrary to return a String[][] representation of the songLibrary or a subset of it from a list of ids.

- Added deleteSong() to saveSongAccessInterface (will need to refactor on Intellij to editSongAccessInterface or something) and to songLibrary.

**KNOWN ISSUES**

- ~Currently no safeguard against artists uploading songs that already exist in the library. I'll likely add a check for songs that have both the same name and same artist as the uploaded song depending on how much I can shave off the runtime of that concept (since imagining the straight forward implementation of this is physically painful).~ Fixed 2022/11/16


src/main/java/Controller/uploadSongController.java
```diff
@@ -1,11 +1,25 @@
...
+import InputData.uploadSongInputData;
+import OutputBoundary.uploadSongOutputBoundary;
+import UseCase.UploadSongInteractor;
 
 public class uploadSongController {
 
-    uploadSongInputBoundary uploadSongInteractor;
-    public void upload(uploadSongInputBoundary uploadSongInteractor){
-        //TODO
+    uploadSongOutputBoundary uploadSongPresenter;
```
sarxhlwalker, line 9: we should decide whether we're actually still passing the presenter into the controller so that it can construct the controller -- i don't think you need to do anything about it for now, but just marking it for future discussion as this has come up on other PRs.

src/main/java/Controller/uploadSongController.java
```diff
@@ -1,11 +1,25 @@
...
+    uploadSongInputBoundary uploadSong;
+
+    public uploadSongController(uploadSongOutputBoundary uploadSongPresenter){
+        this.uploadSongPresenter = uploadSongPresenter;
+        uploadSong = new UploadSongInteractor(this.uploadSongPresenter);
```
sarxhlwalker, line 14: lol could u add this. to uploadSong plz

src/main/java/InputData/uploadSongInputData.java
```diff
@@ -2,7 +2,17 @@
 
 public class uploadSongInputData {
     String filepath;
-    uploadSongInputData(String filepath){
+    String uploader;
```
sarxhlwalker, line 5: is the uploader the artist user's name? could you add a comment explaining what this string refers to/is used for?

src/main/java/ViewModel/uploadSongViewModel.java
```diff
@@ -1,4 +1,33 @@
...
+    }
+
+    /**
+     *
+     * @param filepath Filepath to the desired song.
+     * @param user The username of the uploading user.
+     */
+    public void upload(String filepath, String user){
+        uploadSongInputData inputdata = new uploadSongInputData(filepath, user);
+        uploadSongController.upload(inputdata);
```
sarxhlwalker, line 22: might have to return isUploaded to the View -- not sure what your implementation plans were for the View

sarxhlwalker (dismissed):
Overall looks good. Left some comments, but nothing major.

src/main/java/Controller/uploadSongController.java
chrstinalin: lol oops

src/main/java/InputData/uploadSongInputData.java
chrstinalin: Yeah, username. It'd be used on my end for determining all the songs a user uploaded.

src/main/java/ViewModel/uploadSongViewModel.java
chrstinalin: Ah true. Haven't looked at view yet, probably why i didn't write it.

rafaelgacesa (approved):
I love code review so much. In fact, this pull request made my day! It has tests. It has interfaces that follow the interface segregation principle of SOLID. And it has a use case! It even has painful merge conflicts - the whole version control experience! What more could you possibly want!!

alsorafaelgacesa (approved):

alsorafaelgacesa (approved):

notrafaelgacesa (approved):
Collaborative code review is an imperative part of the software design process. Getting more than one pair of eyes on the code can save hours of debugging!

PR #63, opened 2022-11-17
Recommend song use case

rafaelgacesa:
### In this pull:
- The recommend song interactor, input boundary, controller code, view model code
- Some changes to the space use cases to make an edge case work properly
   - when someone adds a song to the space while it is playing


NicholasMacasaet (approved):
As far as I can tell, save for some minor differences in how you structured your controller compared to mine, the "flow" of information is correct for this use case.

src/main/java/Controller/musicEngineController.java
```diff
@@ -17,13 +17,12 @@ public class musicEngineController {
     private final int PLAYLIST = 1;
     private final int SPACE = 2;
```
chrstinalin, line 18: Is there a particular reason why we're giving each type a int representation rather than just using Strings?

Mistryr222 (approved):
Good implementation of this use case!

src/main/java/UseCase/recommendSong.java
```diff
@@ -1,4 +1,55 @@
...
+        Song randomSong = playlist.get(random.nextInt(playlist.size()));
+        String genre = randomSong.getGenre();
+        Song rec = null;
+
+        for (songDsData song : library.getLibrary()){
+            if (rec == null){
+                rec = song.getSong();
+            }
+            if (genre.equals(song.getGenre()) & !playlist.contains(song.getSong())) {
+                rec = song.getSong();
```
chrstinalin, line 47: Couldn't you just do --

```
rec = song.getSong();
if(genre.equals(rec.getGenre()) && ...) break;
```

?

chrstinalin (commented):
See Comments

src/main/java/UseCase/recommendSong.java
chrstinalin: There's also the issue of iterating through the entire songLibrary and not finding a song with the same genre. Were you just gonna do the very last song it grabs from songLibrary? It's fine if you are, just wondering, lol

src/main/java/Controller/musicEngineController.java
rafaelgacesa: Cleaner to use constants for comparison + comparing strings is kinda futile in general

src/main/java/UseCase/recommendSong.java
rafaelgacesa: As of right now, if no song is found then it will choose the very first song (the only iteration where rec is null). To address the first point - yeah I could do that _if_ I skipped doing the null check. I figured it would be kind of silly to have two variables iterating through the list (both rec and song, if I remove the null check), which is why I chose for rec to hold the first song in the library until a proper recommendation is found.

chrstinalin: Ah, I see now what you were trying to do. To be honest, it feels like they have the number of iterations and my suggestion might be clearer to understand for anyone else who'd have to take a look in the future(at least from my initial outside perspective).

I'm fine with not changing it since its functionally the same if you don't want to though.

chrstinalin (approved):

PR #64, opened 2022-11-17
Completing the audio system use cases

rafaelgacesa:
### In this pull:
- Removed the vestigial `playAudio` related files
   - I thought these were gone already?
- implemented the interactor, boundary, controller, and view model code for pausing a song
- implemented the interactor, boundary, controller, presenter, and view model code for skipping a song
- modified some supporting files minimally to aid these use cases (eg. `MusicPlayer`, `songInputData`)


src/main/java/Controller/musicEngineController.java
```diff
@@ -107,6 +111,18 @@ public void playRecommendation(playlistInputData data){
     }
 
     /**
+     * Skips the current song, if a playlist is playing.
```
sarxhlwalker, line 114: Wouldn't hurt to add a quick explanation why we only care if a playlist is playing (ex. for the TAs LOL).

src/main/java/Controller/musicEngineController.java
```diff
@@ -115,7 +131,7 @@ private void stop(){
             case SONG:
                 break;
             case PLAYLIST:
-                playPlaylist.stopQueue();
+                this.playPlaylist.stopQueue();
```
sarxhlwalker, line 134: in what scenarios would you stop playing the playlist? we should have a reset playbar function to call here (and possibly case SONG) to set the playbar to blank when we're not playing anything.

src/main/java/InputData/songInputData.java
```diff
@@ -12,6 +14,11 @@ public songInputData(Song song){
         this.song = song;
     }
 
+    public songInputData(int id){
```
sarxhlwalker, line 17: I like this implementation. It'll come in handy for addToSpace. Thanks!

src/main/java/UseCase/NextSong.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public void skipSong() {
+        MusicPlayer mp = MusicPlayer.getInstance();
+        int id;
+        if (mp.getCurrentSong() != null)
+            id = playlist.indexOf(mp.getCurrentSong());
+        else {
+            mp.close();
+            presenter.songPlayed(null);
```
sarxhlwalker, line 31: Could you add some comments explaining your logic here? I'm not sure I follow what/why the else branch is doing.

src/main/java/UseCase/NextSong.java
```diff
@@ -0,0 +1,39 @@
...
+        this.presenter = presenter;
+        this.name = data.getName();
+    }
+
+    @Override
+    public void skipSong() {
+        MusicPlayer mp = MusicPlayer.getInstance();
+        int id;
+        if (mp.getCurrentSong() != null)
+            id = playlist.indexOf(mp.getCurrentSong());
```
sarxhlwalker, line 28: if you call the presenter in the else branch, why aren't you calling the presenter here? if it's handled in play.play(), could you add a comment saying so?

src/main/java/ViewModel/musicEngineControllerViewModel.java
```diff
@@ -31,6 +32,21 @@ public void getRecommendationAction(int id){
...
+    public void pauseSongAction(){
+        this.musicEngineController.pauseSong();
+    }
+
+    public void skipSongAction(){
+        this.musicEngineController.playNext();
+    }
+
+    public String playSongAction(int id){
+        songInputData data = new songInputData(id);
```
sarxhlwalker, line 44: IIRC we were creating the InputData in the Controller. Technically, this does not violate clean architecture but I would prefer that we stay consistent. This may be another thing we should discuss (along with where we are constructing the interactor -- the ViewModel or the Controller) and we might want to open an Issue for refactoring this later once all of our use cases are functional.

sarxhlwalker (approved):
No major complaints. I left some comments, but they aren't urgent.

src/main/java/Controller/musicEngineController.java
rafaelgacesa: `stop()` is a bit of a misleading name, it doesn't immediately stop the currently playing song, it stops the queuing of future songs. thus, any time you want to play something new (eg. a song, another playlist, a space), the playlist must be stopped.

src/main/java/Controller/musicEngineController.java
rafaelgacesa: fixed

src/main/java/UseCase/NextSong.java
rafaelgacesa: I commented the index of line, but I hesitate to add one to the else branch to avoid a code smell. basically, the else branch (which I've now slightly modified to check another condition - if the index is the last one in the playlist) only occurs if skipSong() is somehow called when there is no playlist playing, so it just closes the media player, and returns null to the presenter which I will handle later when I implement the play bar.

src/main/java/UseCase/NextSong.java
rafaelgacesa: fixed

src/main/java/ViewModel/musicEngineControllerViewModel.java
rafaelgacesa: fixed

sarxhlwalker (approved):

NicholasMacasaet (approved):
From what I can see, all the logic and code flows very nicely, well done.

Mistryr222 (approved):
Good changes made!

sarxhlwalker (approved):
cool

PR #67, opened 2022-11-18
Views!

Mistryr222:
1. Added templates to all views - test your work by creating a frame in main
2. Added Swing functionalitiy to all views with comments for each persons use case 
3. Everyone else will need to add their use cases and some Swing code, follow the syntax, notes, and if you have any questions just ask me

### DIFF
--- a/src/main/java/View/artistView.java
+++ b/src/main/java/View/artistView.java
@@ -1,4 +1,87 @@
 package View;
 
-public class artistView {
+import javax.swing.*;
+import java.awt.*;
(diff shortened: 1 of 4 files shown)


NicholasMacasaet (approved):
All of the comments for directing everyone to where they need to code are very good. I don't see any problems with this.

rafaelgacesa (approved):
Great templates. Hopefully this means we have something functional for Monday.

chrstinalin (approved):
Looks good, thanks for doing this!

PR #70, opened 2022-11-18
Javadocs!!

rafaelgacesa:
### Discuss this pull at meeting on the 18th

You are free to read through the code in this pull if you really want to, but that might even be more painful than it was to write. **There are no changes to any program logic in this pull.**
### In this pull:
- Added javadocs pretty much everywhere they were missing
- removing vestigial files
   - some random empty classes that we weren't going to use ever due to changes in logic
- addressed unused import warnings
- addressed modifier warnings (public, final, etc.)

**This may be difficult to merge due to conflicts. We can either:**
1. not merge it at all
2. merge it and then have everyone merge main into their branch (this may also lead to conflicts, however we can just ignore them and hopefully it doesn't cause too much of a mess)


NicholasMacasaet (dismissed):

chrstinalin (dismissed):

rafaelgacesa:
This should be one of the last things that is merged - please don't merge it even if it has 3 reviews

PR #75, opened 2022-11-19
Feature 6 search engine Views 

manya-mittal:
This pull request contains quite a lot of changes I made for the Search Use Case. I have included a summary below. 

- The main change is the searchOutputView which uses Swing to create the JTable of the search results. This is the facade design pattern as the complex frontend Swing code is hidden in the  searchOutputView class, rather than being in Main. 
- I also added a search bar function in playListView, as well as the functionality of being able to switch between screens and changing the output based on what they user types
- Adjusted the way I was accessing songLibrary to match the singleton design pattern that Christina used. 
- Modified some files to adhere to clean architecture 
- Refactor some methods to make helper functions and remove unnecessary comments to reduce code smells


rafaelgacesa (approved):
Looks great! View code/use case integrated into the view work really nicely.

sarxhlwalker (approved):
looks good

SiqiZhang6 (approved):
looks great! thanks for the work

PR #77, opened 2022-11-19
Feature 10 add to space

sarxhlwalker:
popups for add song to space

this pr got a bit larger than intended, so some additional comments: 
- fixed bug where you could click listen to space when you're currently listening to it and it'd shuffle
- fixed bug regarding queueing of space songs
- did the view aspect for add song to space.
- fixed the issue that the space data would not persist
- fixed some typing issues (ex. arraylist -> list)


src/main/java/Controller/musicEngineController.java
```diff
@@ -59,21 +58,23 @@ public void playSpace(){
...
-     * @param songInputData input data for adding a song to the space
+     * @param songToAddID id of song to add to space
      */
-    public void spaceAddSong(songInputData songInputData){
-        int songToAddID = songInputData.getSong().getID();
+    public void spaceAddSong(int songToAddID){
         for (songInputData currSongInputData : this.spaceSongList){
             int currSongID = currSongInputData.getSong().getID();
             if (currSongID == songToAddID){
+                this.spacePresenter.notAddedToSpace(currSongInputData.getName());
```
manya-mittal, line 67: This seems like it could be use case code which might be better in a use case interactor, but we don't have an addToSpaceInteractor. I don't think this is an issue because of the way we are dealing with spaces. If it was a singleton, it might've made sense to create an interactor for it but since we are using space as a List of songs in the controller (as per the profs reccomendation), it doesn't make sense to have to keep passing around the List of songs. Although it's not typical for the controller to talk to to presenter, I think it's fine in this case since they're both in the same layer so it shouldn't be a violation of clean architecture.

src/main/java/Main.java
```diff
@@ -24,7 +25,8 @@ public static void main(String[] args){
 
 //        // TODO
         User user = new RegularUser("User", "");
-        new playlistView(user);
+        playlistView PlaylistView = new playlistView(user);
```
manya-mittal, line 28: This is still using Users in Main which is a violation of clean architecture I believe since Regular User is an entity. I know it's just temporary to make the Views work but we should probably merge this pull request before the one that @rafaelgacesa made since his PR deals with this issue with InMemoryUsers. Merging this pull request first will probably avoid unnecessary merge conflicts, since the other pr will delete this anyways.

src/main/java/UseCase/Search.java
```diff
@@ -17,8 +17,10 @@ public class Search implements searchInputBoundary {
     private final searchOutputBoundary searchPresenter;
 
        private GetSongAccessInterface songLibrary;
+    // TODO: Deal with singleton
 
     public Search(searchOutputBoundary searchPresenter){
+        // import songLibrary instead of passing it in
```
manya-mittal, line 23: hasn't this already been imported in through GetSongAccessInterface?

src/main/java/View/searchOutputView.java
```diff
@@ -12,37 +10,32 @@
...
+    public searchOutputView(String searchText, playlistView playlistView){
+        this.initialiseValues(searchText, playlistView);
         this.setUpTable();
-        this.initialiseFrame();
+        this.initializeFrame();
     }
 
-    public void initialiseValues(String searchText){
+    public void initialiseValues(String searchText, playlistView playlistView){
+        this.playlistView = playlistView;
```
manya-mittal, line 38: Is this necessary? I don't think it will be needed once we get the logged in users working, since we can just retrieve all the information again. I believe raf's pull request will deal with this. It will cause conflicts if we merge your pull request that has playlistView as a parameter and then try to merge rafs after which doesn't have it. I think it might just be easier to delete it from yours before we merge this

src/main/java/View/playlistView.java
```diff
@@ -55,40 +52,32 @@ public playlistView(User user){
...
             String searchText = this.searchBar.getText();
             this.searchViewModel.search(searchText);
-
-            this.jframe.dispose();
-            new searchOutputView(searchText);
-
-        } // TODO -- NOTE: add your action commands as an else-if to this if statement
-
+            System.out.println(searchText);
+            new searchOutputView(searchText, this);
```
manya-mittal, line 75: same comment as what I wrote for searchOutputView. Passing in the playlistView instance is not needed for searchOutput view and will avoid conflicts when merging raf's pull request as well.

manya-mittal (changes requested):
This mostly looks good, I just requested a few changes in searchOutputView to avoid merge conflicts later when raf's pull request is also merged. After a few minor changes this pull request should be good to go!

manya-mittal (dismissed):
This looks good to me now! Thanks for making the changes I asked and for merging with rafs branch first to avoid the conflicts.

src/main/java/Controller/musicEngineController.java
SiqiZhang6: To expand a bit on prof's recommendation, we got rid of the addToSpace use case as a whole because adding song to space does not require reaching to the back end. (because there is no specific rule we're just adding to the list). And yes I agree that this does not violate clean architecture because the presenter and the controller are in the same layer and thus allowed to talk to each other.

src/main/java/UseCase/RegisterArtistInteractor.java
```diff
@@ -16,7 +16,7 @@ public class RegisterArtistInteractor implements RegisterArtistInputBoundary{
 
     public RegisterArtistInteractor(loginOutputBoundary registerPresenter){
         this.allUsers = Database.userList.getInstance();
-        this.registerPresenter =registerPresenter;
+        this.registerPresenter = registerPresenter;
```
rafaelgacesa, line 19: Can you undo this change? I made the same one in Jessica's open PR but this will cause merge conflicts.

sarxhlwalker: lmaooo sure

src/main/java/UseCase/Search.java
sarxhlwalker: oops yeah i think i accidentally copied that in when dealing w merge conflicts

rafaelgacesa (approved):
Thanks for fixing the potential conflicts. Looks good.

manya-mittal (approved):
looks good!

chrstinalin (approved):

PR #78, opened 2022-11-19
Playlist view, playbar, user login view

rafaelgacesa:
I apologize for the breadth and size of this pull - I should have split it into further branches but I was so deep in that it was too late.

### In this pull:
- implementation of data structures in View layer and their respective interface in the ViewModel layer
   - `UserDsView`/`InMemoryUser`, `SongDsView`/`InMemorySong`, `PlaylistDsView`/`InMemoryPlaylist`
   - these are used to pass data around from different views and from the view model to the views
   - this required a small addition to `playlistInputData` to function
   - try following the logic from `loginRegisterView` if you are confused about how to pass data with these objects
- updated `playlistView` to use `InMemoryUser` rather than `User`
- the main dashboard playlist view is now built up in `PlaylistPanelView`, and then added to the active view in `playlistView`
   - uses the `InMemoryPlaylist` objects in the `InMemoryUser`'s list of playlists to load in elements of each playlist and display them
   - then the constructed panel is grabbed from `PlaylistPanelView` by `playlistView` and added to the main frame
- the play bar is now built up in `PlayBar`, and then added to the active view in `playlistView`
   - similar to `PlaylistPanelView`
   - added logic for synchronization to update play bar in `MusicPlayer`, `MusicEngineControllerViewModel`, and `PlayBar`
- completed rough login use case, and implemented `loginRegisterView`
   - completed `Login`, `loginController`, `loginInputBoundary`, `loginOutputBoundary`, `loginOutputData`, `userPresenter`, `userViewModel` such that logging in as a regular user now works
   - had to modify `registerInteractor` to avoid program crash, but no logic changes made
   - **artist login and all registration is not implemented yet**, please read and change what is necessary @SiqiZhang6 
   - required a small change to `userList` since it was throwing a null pointer exception when checking password
- updated `searchOutputView` to work with a `InMemoryUser` rather than a `User`
   - please read over this @manya-mittal, I also implemented the 'add to playlist' where it displays all of the user's playlist (but not the button actions, that's for @NicholasMacasaet)
- fixed a bunch of bugs in `nextSong` use case
   - required changes to `musicEngineController`, `nextSong`, `nextSongInputBoundary`, `playPlaylist`, `PPlaylistTest`, and `musicEngineControllerViewModel`
- fixed a song duplication bug in `playlistDsData`

If you run `Main` as is in this pull, you should get a functioning program! Also, I ask that we leave `users.csv` and `playlists.csv` as they are in this pull for now to avoid bugs like the ones encountered in @NicholasMacasaet's branch.

I believe I have mentioned every file modified in this pull here, so hopefully that helps you if you're reviewing the code. Lmk if I missed anything.


NicholasMacasaet (approved):

chrstinalin (approved):
Looks good, hopefully we can get this in to make the proper changes with the InMemory stuff before submitting today.

SiqiZhang6 (approved):
looks great to me, I'll write up the other login/register use cases once this is merged. Thanks for your work

src/main/java/UseCase/Login.java
```diff
@@ -1,21 +1,38 @@
...
+        this.loginUsers = Database.userList.getInstance();
+        this.getUsers = Database.userList.getInstance();
         this.loginPresenter = loginPresenter;
     }
 
     @Override
-    public void login(loginInputData loginID) {
-        //TODO: currently implementing
+    public boolean login(loginInputData loginID) {
+        if (!loginID.isArtist()){
```
SiqiZhang6, line 23: I believe we don't need to check whether loginID is an artist because login use case is the same for both type of users, the difference only occurs in the view, which is why we will take care of that in the view layer instead of here to avoid duplicate code

src/main/java/View/InMemoryUser.java
```diff
@@ -0,0 +1,38 @@
+package View;
+
+import ViewModel.PlaylistDsView;
+import ViewModel.UserDsView;
+
+import java.util.ArrayList;
+
+public class InMemoryUser implements UserDsView {
```
SiqiZhang6, line 8: This solves the issue we had with view unable to access user information since that would violate the clean architecture. Originally we planned to put these data in the loginoutputData, but having this class would work better. Thank you for making this.

PR #79, opened 2022-11-19
ArtistView, songLibrary and uploadSong use case tests, other bits and pieces

chrstinalin:
CHANGES
- Barebones implementation of the ArtistView.
- Fixed admin in artists.csv
- Removed removeSong for the sake of time due to the number of issues it will cause with upload that I don't have time to fix, currently (might go back and do it, probably not).
- Minor fixes to songLibrary. saveFile is now saveLibrary.
- Added uploadQueue to songLibrary. saveLibrary will copy all files in queue to the songLib folder and update songLibrary accordingly.
- Fixed SLibraryTest, implemented UploadSongTest.
- Added an InMemoryArtistUser.
- **Added the main view, LaunchView.**

image
image

KNOWN ISSUES
- **On window close, songLibrary's saveLibrary() needs to get called so that songs.csv and the songLib folder is updated.**
- Uploaded songs currently do not appear in the Jtable. I hate 2D arrays and I refuse to write 500 lines of code just to accommodate it. I'm going to look into using the Vector version of JTable after Monday.
- The ID column is not hidden.  Arguably not the worst thing for the artist end, but should probably change it.
- ~I need to update tests lol~ Fixed 2022/11/21
- Need to replace User instance in view as per #78


rafaelgacesa (commented):
Will take a closer look later, but mostly looks good and most changes are to nicely isolated files. 

Thought it might be worth mentioning that whenever https://github.com/CSC207-2022F-UofT/course-project-team-communify/pull/78 merges, you’ll want to look at using an InMemoryUser object rather than an ArtistUser in your view.

rafaelgacesa (approved):
Looks good!

sarxhlwalker (approved):
wow stellar!

manya-mittal (approved):
looks great!

PR #83, opened 2022-11-21
Feature 4 playlist creation ver2

NicholasMacasaet:
@rafaelgacesa when you get time could you take a look at the bug w. InMemoryUser ?


src/main/java/Database/playlistLibrary.java
```diff
@@ -124,7 +124,9 @@ public playlistDsData findPlaylist(int id) {
             return new playlistDsData(p);
         }
 
-        return playlistDatabase.get(id);
```
chrstinalin, line 127: What you had originally here was fine -- hashmap's default return value if id is not in the keys is null

src/main/java/View/InMemorySong.java
```diff
@@ -17,7 +17,7 @@ public class InMemorySong implements SongDsView {
      * Empty constructor.
      */
     public InMemorySong(){
-
+        this.name = "";
```
chrstinalin, line 20: Remove?

chrstinalin (approved):
Don't see any glaring issues 👍

src/main/java/View/InMemorySong.java
rafaelgacesa: Necessary to avoid null pointer exception

src/main/java/Database/playlistLibrary.java
rafaelgacesa: Not sure if that’s true, I was getting a NullPointerException trying to get playlists that didn’t exist

chrstinalin: Damn, really? If I remember right my implementation of SongLibrary relied on the null return and from what I saw (via good ol print checks) it /was/ indeed returning them. That's weird

rafaelgacesa: It could just be me being stupid lol but this change isn’t breaking anything so might as well just leave it.

sarxhlwalker (approved):
i assume this should be merged after #89 but otherwise looks fine to me

rafaelgacesa (approved):
Looks good

PR #87, opened 2022-11-21
songLibrary Persistence

chrstinalin:
Added windowClosingListener to save songLibrary once window is closed.


sarxhlwalker (approved):
looks good to me. if you want you can add a comment in windowClosing() to explain why it's OK to call save library in the view (cause both in blue layer), but not necessary.

sarxhlwalker (approved):
comment good nice indeed.

rafaelgacesa (approved):
Nice! Good use of listeners.

SiqiZhang6 (approved):
awesome tysm

PR #88, opened 2022-11-21
Moving Input Data creation to controller

manya-mittal:
This is super quick - I just moved where I create the input data. It was previously in the view model but I have now moved it to the controller to be consistent with everyone else


chrstinalin (approved):
Yay consistency

rafaelgacesa (approved):
Sure

sarxhlwalker (approved):
looks great, thanks for this!

PR #91, opened 2022-11-24
View Design Overhaul #1

chrstinalin:
CHANGES

- Basic overhaul of the design and layout of most views. 
- songLibrary edits to accommodate new covers
- Created button icon assets.
- Play Space button no longer uses the String returned by callPlaySpace() @sarxhlwalker as FYI

VERSION 1:

image

preview

VERSION 2 (CURRENT): 

GIF 2022-11-26 7-53-55 PM

STILL NEED TO DO:

- ArtistView, Search fix
- ~Make PlaylistPanel less condensed -- More padding around the title, fake playlist cover from a random song, replace buttons.~ Done 2022/11/24
- ~Lock pause/skip button location to centre of the screen. Prevent pause/skip buttons from moving with song changes.~ Fixed 2022/11/24
- ~Replace dashboard buttons with icons once finalized (and hopefully once I find the ones Flatlaf seems to come with...)~ Fixed 2022/11/24
- ~Update covers so they're not all default ones.~

OPTIONAL:

- ~Making the playbar visible on the search view.~
- Change text fields so you don't need to delete the text in it to log in (Swing doesn't actually have a way to do this, fun.)
- Calculate the formulas to move elements depending on whether 'is artist' is ticked on the login/register screen.

KNOWN ISSUES:

- Song retrieval on launch: facing performance issues when actions are interrupted by it (ex. pressing back from login). **VERY IMPORTANT AS OF COVER UPDATE**
- ~Song controls disappear when Space is played~
- ~'Play' icon does not update when playlist is played~
- ~Skip button disappears if you play a song from Search view~
- ~Home in search view throws a NullPointerException~
- Names get cut off in playlistPanelView.
image


src/main/java/View/playlistView.java
```diff
@@ -22,19 +22,31 @@ public class playlistView extends JFrame implements ActionListener {
...
+    private int DEFAULT_HEIGHT = 40;
+    private int DEFAULT_KERNING = 20;
+
+
     /**
      * constructor
      * @param user takes in the user's data to display their own dashboard!
      */
-    public playlistView(InMemoryUser user){
+    public playlistView(InMemoryUser user, ImageIcon icon, ImageIcon logoImg, ImageIcon logoSmall){
```
rafaelgacesa, line 46: The null pointer exception thrown due to the home button in search is due to this change here - search output view uses the second overloaded constructor which does not have these icon parameters - then presumably you try to access the icon objects but they don't exist in playlistView objects made by searchOutputView.

chrstinalin: Hm, okay. I'll fix that later then. I'm thinking of changing it slightly anyway

rafaelgacesa:
> * Lock pause/skip button location to centre of the screen. Prevent pause/skip buttons from moving with song changes.
I wish you serious luck with this... I spent a good ~hour trying to do this to no avail. (the only 'viable' solutions to this are probably learning how to use Swing's `GridBagLayout` which is not worth your time, or to hard-code the position of each element, which is gross.)

> * Making the playbar visible on the search view.
This should be as easy as adding the object to the main JFrame, since the search view already stores the playbar.

> * Names get cut off in playlistPanelView.
This is probably just because you changed the resolution of the main JFrame, so it should be an easy fix.

chrstinalin:
LOL ITS SO SLOW AUTOGRADER RUNS OUT OF MEMORY. okay i have to fix that.

chrstinalin:
To document:

Shrunk album covers by ~1200%, fixing memory test issue. New issue of playbar tests possibly relying on different song objects now, or some other issue? Have to take a closer look tbh

image

sarxhlwalker (dismissed):
cute!

sarxhlwalker (approved):
looks good, thanks for fixing the home button bug

rafaelgacesa (approved):
I love this code. It is functional and beautiful. I will sleep well knowing that this code is in our codebase. Thank you for your contributions.

rafaelgacesa (approved):
I hate you. My time is valuable. Never do this again.

SiqiZhang6 (approved):
Thank you so much for making this change! it looks a lot better now :D

PR #105, opened 2022-12-03
Tests !

NicholasMacasaet:
Some tests added for playlist view model for each of my use cases.


chrstinalin (approved):
view model tests 😵

SiqiZhang6 (approved):
the tests looks great, thx

sarxhlwalker (approved):
looks good to me. methods should typically start with a lower case letter (not capital, as in your test file), but that'll be fixed in refactoring.

PR #106, opened 2022-12-04
Refactoring

sarxhlwalker:
TL;DR OF MY CHANGES: 
* renamed packages to start with lower case letters
* made everything final that intellij wanted to be final
* fixed SO MANY typos
* made duplicated code private helper functions where possible
* removed unused overridden constructors
* removed unused parameters
* removed Presenter instance variables from Controllers where they were just stored and not used anywhere (i.e. saved as instance variable only to be passed to interactor and then never mentioned again) 
* renamed some test files (now they are all singular instead of half plural) and made test methods start with lower case letters

Some things I've noticed: 
* @chrstinalin SongLibrary.java line 35 — private instance variable of ArrayList. should this be List instead? I forget if that's how Liskov works. 
* @NicholasMacasaet PlaylistController (in the Interface Adapter layer) imports a Song Entity (in the Enterprise Business Rules layer) — this should be illegal wrt clean architecture. i’m not confident that i can fix this without screwing everything up. could you take a look? 
* @chrstinalin @SiqiZhang6 (I assume it's one of you two?) there's duplicated code between ArtistView and LaunchView — not sure how to fix since it's going across two different classes. thoughts? 

OTHER NOTES: 
* it wants us to uncomment some assertion statements :(
* lots of unused methods 
* from the some-300 warnings from friday, we're now sitting at a hot 37 warnings + 7 weak warnings = 44 warnings. (For some context, "Unused declarations" takes up 27/44 of those. I don't want to blindly mass delete, though, so I'd like to have everyone together and go over it on Monday.)


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,15 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
```
rafaelgacesa, line 3: Pls delete this file

rafaelgacesa (approved):
thank u for doing the stuff nobody wants to do lol. looks good to me now, very grateful much appreciated

NicholasMacasaet (approved):
Looks good, although I dont see what you mean by importing a song entity in any of my controller classes (I could have missed it) those classes only take in a song ID as an input for their methods, also playlistcontroller is implemented by Rohan

sarxhlwalker:
> Looks good, although I dont see what you mean by importing a song entity in any of my controller classes (I could have missed it) those classes only take in a song ID as an input for their methods, also playlistcontroller is implemented by Rohan @sarxhlwalker

oh mb, didn't realize that. for reference, this is what i'm referring to (@Mistryr222):
Screenshot 2022-12-04 at 12 52 03 AM

Mistryr222 (approved):
Nice catch on the PlaylistController having a song entity so this will change the playlistInputData as it will need to cast (Song) in its for loop, I have made changes on my end we can review them on Monday but the remaining refactoring is excellent. Thank you!

PR #110, opened 2022-12-06
The Final Pull Request 🫡 (refactoring, tests, final changes)

### LINKED ISSUE #102: fix intellij warnings

### LINKED ISSUE #104: ensuring imports follow clean architecture
a number of entities (most often songs) were imported into other layers, like controllers, database, presenters. 

a number of inner layers also import view.inmemory classes, which i'm less certain about whether that's allowed.

rafaelgacesa:
This is (hopefully) the last code I write and the final pull request made before submission!
The autograder will fail since the tests are not designed to be run all at once (issues with `MusicPlayer` entity and the `isPlaying()` method when being pulled and tugged from all the different tests at the same time).
I ran and tested most, if not all of the features and they are working as intended.

### In this pull:
- down to 10 'non-removable' warnings
- uncommented all assertions and modified tests
- prepared User1 to look better
   - changed playlists, removed all the test playlists
- modified ~10 classes to adhere with clean architecture
  - mostly pretty straightforward changes to get rid of bad imports, some classes required a bit more extensive work
- the view is no longer slow 🐢


rafaelgacesa:
> Could you write out (briefly) what are the remaining 10 'non-removable' warnings?

4 ‘suspicious calls to arraylist.contains’ in the view layer since we are not casting buttons to IDButtons/DoubleIDButtons (what the arraylists contain), which we can’t change since there’s two different types of objects. 

4 ‘parameter is always the same’ warnings in the singleton databases since the paths don’t change and the objects are only constructed once, but I’d rather we don’t hardcode paths as instance variables. 

2 ‘unused declaration’ warnings for SaveSongAcessInterface since there’s two methods that aren’t used in the actual program, but they are used in tests so I don’t want to remove them. 

There’s also a ‘weak warning’ about the ‘outdated’ version of JUnit we’re using, but we can’t update it without breaking tests.

sarxhlwalker (dismissed):
thanks for making these much-needed changes

sarxhlwalker (dismissed):
thanks for fixing these issues, raf

sarxhlwalker (approved):
lgtm

chrstinalin (approved):
🫡

NicholasMacasaet (approved):
🫡

